PR #1, opened 2022-09-06
Use latest code from scipp main

nvaytet:


tests/graph_nodes_test.py
```diff
@@ -2,8 +2,8 @@
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
 
 import scipp as sc
-from plopp import Plot, Figure, widgets, input_node, node
-from plopp.widgets import widget_node
+from scipp.experimental.plotting import Plot, Figure, widgets, input_node, node
+from scipp.experimental.plotting.widgets import widget_node
```
SimonHeybrock, line 6: Why is this not in `plopp`?

nvaytet: nice catch!

PR #7, opened 2022-09-06
Add docs skeleton

nvaytet:
Add framework for making docs (conf.py, index.rst), and a short example notebook, to make sure everything is set up properly.

More docs will follow in future PRs.


src/plopp/__init__.py
```diff
@@ -16,7 +16,7 @@
 from .graph import show_graph
 from .plot import Plot
 from .model import Node, node, input_node
-from .figure import Figure
+from .figure import figure
```
SimonHeybrock, line 19: Will function name matching file name cause trouble?

nvaytet: Good point. I don't remember in which case it does and doesn't.
For the sake of safety, I will change the file name.

PR #9, opened 2022-09-08
Remove hard dependency on ipywidgets and make SVG figures

nvaytet:
- Remove hard dependency on ipywidgets
- make SVG static figures using `_repr_mimebundle_`
- Separate toolbar from figure better
- Add functions on toolbar and figure, such as you can call `home` or `logx` on the toolbar or the figure and they will update each other


src/plopp/io.py
```diff
@@ -0,0 +1,14 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
+
+from io import BytesIO
+
+
+def fig_to_pngbytes(fig, form='png'):
```
SimonHeybrock, line 7:
```suggestion
def fig_to_bytes(fig, form='png'):
```
?

src/plopp/io.py
```diff
@@ -0,0 +1,14 @@
...
+    """
+    Convert a Matplotlib figure to png bytes.
```
SimonHeybrock, line 9: Needs update?

src/plopp/static.py
```diff
@@ -0,0 +1,37 @@
...
+
+class StaticFig(Figure):
+
+    def _repr_mimebundle_(self, include=None, exclude=None):
+        """
+        Mimebundle display representation for jupyter notebooks.
+        """
+        return {
+            'text/plain': 'Figure',
+            'image/svg+xml': fig_to_pngbytes(self._fig, form='svg').decode()
```
SimonHeybrock, line 16: Is the size of 2d figures manageable? I expect an svg is much larger than png in that case?

SimonHeybrock (approved):

src/plopp/static.py
nvaytet: yes it is, but now we `rasterize` the `pcolormesh`

SimonHeybrock: Meaning... the mesh is not vector, but all the rest is?

PR #12, opened 2022-09-14
Add safety net when trying to plot very large data

### LINKED ISSUE #2: Set default limit on shape to avoid freezes/crashes when plotting large objects

nvaytet:
Limits are:
- 1d plots: 1M points (figure creation creation=3s, zoom=<1s)
- 2d plots: 2500x2500 image (figure creation=4s, zoom=1.5s)

Limits are only applied when using the `plot` wrapper, not when using `figure` directly.

The size check can be disabled using `plot(da, ignore_size=True)`

Fixes #2


src/plopp/prep.py
```diff
@@ -27,7 +27,16 @@ def _convert_if_not_none(x, unit):
     return x
 
 
-def preprocess(obj, crop=None, name=''):
+def _check_input_data_size(obj):
+    limits = {1: 1_000_000, 2: 5000 * 5000}
```
SimonHeybrock, line 31: Regarding your question, I think we should determine limits based on approximate timings of creating such a plot.

src/plopp/prep.py
```diff
@@ -27,7 +27,16 @@ def _convert_if_not_none(x, unit):
...
+    if obj.ndim not in limits:
+        raise ValueError("plot can only handle 1d and 2d data.")
+    if prod(obj.shape) > limits[obj.ndim]:
+        raise ValueError("Attempting to plot data that exceeds reasonable limits. "
+                         "To disable this check, use `ignore_size=True`")
```
SimonHeybrock, line 36:
```suggestion
        raise ValueError(f"Plotting data of size {size} may take very long or use an excessive amount of memory. "
                         "This is therefore disabled by default."
                         "To disable this check, use `ignore_size=True`")
```
or something like that?

src/plopp/prep.py
nvaytet: I think there is both figure creation and then interactions like zooming which can also be slow. I think it's quite hard to determine but we should experiment?

nvaytet: I get the following timings on my workstation:

**2d image:**
- 10000x10000: creation=1min, zoom=20s
- 5000x5000: creation=14s, zoom=5s

**1d marker plot:**
- 1M points: creation=3s, zoom=<1s
- 5M points: creation=14s, zoom=1s
- 10M points: creation=28s, zoom=2s

**1d line plot (no markers):**
- 5M points: creation=2s, zoom=<1s

So should we go with 5M points, and 5000x5000?

SimonHeybrock: I think we should choose the limits such that wait time does not exceed a couple of seconds. 10 seconds is already too long probably.

src/plopp/prep.py
```diff
@@ -31,9 +31,10 @@ def _check_input_data_size(obj):
     limits = {1: 1_000_000, 2: 5000 * 5000}
     if obj.ndim not in limits:
         raise ValueError("plot can only handle 1d and 2d data.")
-    if prod(obj.shape) > limits[obj.ndim]:
-        raise ValueError("Attempting to plot data that exceeds reasonable limits. "
-                         "To disable this check, use `ignore_size=True`")
+    if (size := prod(obj.shape)) > limits[obj.ndim]:
+        raise ValueError(f"Plotting data of size {size} may take very long or use an "
```
SimonHeybrock, line 35: Include shape in error?

src/plopp/prep.py
nvaytet: So i'll go with 1M and 2500x2500

SimonHeybrock (approved):

PR #17, opened 2022-09-15
Plot 2d coords

### LINKED ISSUE #3: Plotting with 2D coords does not work

nvaytet:
```Py
import scipp as sc
import numpy as np
import plopp as pp

# make some data
N = 10
M = 5
x = np.arange(N+1).astype(np.float64)
y = np.arange(M+1).astype(np.float64)
xx, yy = np.meshgrid(x, y[:-1])
z = np.random.random([M, N])
for i in range(M):
    xx[i] *= (i + 1.0)

da = sc.DataArray(data=sc.array(dims=['y', 'x'], values=z, unit='counts'),
 coords={'x': sc.array(dims=['y', 'x'], values=xx, unit='m'),
         'y': sc.array(dims=['y'], values=y, unit='m')})

pp.plot(da)
```
Screenshot at 2022-09-15 16-47-27

Fixes #3


tests/wrappers_test.py
```diff
@@ -71,3 +71,18 @@ def test_plot_rejects_large_input():
...
+    pp.plot(da.transpose())
+
+
+def test_plot_2d_coord_limits():
+    p = pp.plot(dense_data_array(ndim=2, ragged=True))
+    assert np.array_equal(p._ax.get_xlim(), [-0.5, 88.5])
+    assert np.array_equal(p._ax.get_ylim(), [-0.5, 49.5])
+    p = pp.plot(dense_data_array(ndim=2, ragged=True, binedges=True))
+    assert np.array_equal(p._ax.get_xlim(), [0.0, 89.0])
+    assert np.array_equal(p._ax.get_ylim(), [0.0, 50.0])
```
SimonHeybrock, line 88: Tests like this don't follow what we have, e.g., learned from the "Software Engineering @ Google@ book. This is just checking some numbers, it is unclear what they mean (since we do not see the data setup), and probably the test will break if `dense_data_array` changes somehow?

nvaytet: True. So is doing `assert np.array_equal(p._ax.get_xlim(), [da.coords['xx'].min().value, da.coords['xx'].max().value])` better?

nvaytet: On second thought, I think I will just remove this test. It sounds like it's going to break all the time, and it in fact has nothing to do with 2d coords, the problem was there also with 1d coords. It's the bin edges which is important.

src/plopp/mesh.py
```diff
@@ -76,16 +74,49 @@ def __init__(self,
...
+        clip = (slice(1, -1), slice(None))[::1 - (2 * axis)]
+        # Broadcast 1d coord to 2d
+        broadcasted_coord = broadcast(xy[dim_1d],
+                                      sizes={
+                                          **xy[dim_2d].sizes,
+                                          **xy[dim_1d].sizes
+                                      }).transpose(self._data.dims).values
+        out = {
+            'xy'[axis - 1]:
+            np.repeat(broadcasted_coord, 2, axis=axis)[clip],
```
nvaytet, line 109: @SimonHeybrock do you have a good idea on how we could do more of this with scipp instead of numpy? I think the `axis` parameter is a bit bug-prone, would be nice to just use dimension labels...

src/plopp/tools.py
```diff
@@ -27,6 +27,21 @@ def coord_as_bin_edges(da, key, dim=None):
         return concat([left, center, right], dim)
 
 
+def repeat(x, dim, n):
+    """
+    Equivalent of Numpy's `repeat` function: repeat the values of a variable `n` times
+    along dimension `dim`.
+    """
+    index = x.dims.index(dim) + 1
+    dummy_dim = '_'
```
SimonHeybrock, line 36: This is going to break if that matches a dim name, use `uuid.uudi4()` (grep in `scipp` to see how we do this in other places).

src/plopp/mesh.py
```diff
@@ -74,42 +74,52 @@ def __init__(self,
...
-                                          **xy[dim_1d].sizes
-                                      }).transpose(self._data.dims).values
+        # Broadcast 1d coord to 2d and repeat along 1d dim
+        broadcasted_coord = repeat(broadcast(xy[dim_1d[0]],
+                                             sizes={
+                                                 **xy[dim_2d[0]].sizes,
+                                                 **xy[dim_1d[0]].sizes
+                                             }).transpose(self._data.dims),
+                                   dim=dim_1d[1],
+                                   n=2)
```
SimonHeybrock, line 111: Can the `repeat` and `broadcast` be swapped? Is this expensive otherwise?

src/plopp/mesh.py
```diff
@@ -74,42 +74,52 @@ def __init__(self,
...
+        return repeat(self._data.data, dim=dim_1d[1], n=2)[dim_1d[1], :-1].values
 
     def _from_data_array_to_pcolormesh(self):
         xy = {k: coord_as_bin_edges(self._data, self._dims[k]) for k in 'xy'}
         z = self._maybe_repeat_values(self._data.data)
 
-        if self._no_2d_coord():
+        dim_1d, dim_2d = self._get_dims_of_1d_and_2d_coords()
+        if dim_2d is None:
             return xy['x'].values, xy['y'].values, z
```
SimonHeybrock, line 102: Is this avoiding the potentially costly stuff below for "normal" coords?

nvaytet: yes

src/plopp/mesh.py
nvaytet: hmm would not be very easy I think. I can add a TODO note?

SimonHeybrock (approved):

PR #19, opened 2022-09-16
Add dimension slicer plot

### LINKED ISSUE #14: Add slicer plot for slicing higher dimensions

nvaytet:
This allows to slice higher dimensions with sliders.

Example:
```Py
%matplotlib widget
import plopp as pp

da = pp.data.dense_data_array(ndim=3)
pp.slicer(da, keep=['z', 'y'])  # slice out dimension 'x' -> makes a 2d image with 1 slider
pp.slicer(da, keep=['z'])  # slice out dimensions 'x' and 'y' -> makes a 1d line with 2 sliders
```

We are not adding the buttons from the old plotting that allowed to flip the dimensions interactively because this would lead to too much entangled code between the `Mesh` and the `Figure`, where a new `Mesh` would have to be made, the axis limits would have to be changed, and what to do with the `log` scales of the axes.

Instead, we make an easy to use API where `keep` is a `list` of dimensions to allow the user to quickly get what they want by re-plotting, instead of changing dims dynamically after the plot was made.

Fixes #14


SimonHeybrock:
> ```python
> pp.slicer(da, dims=['x'])  # slice out dimension 'x' -> makes a 2d image with 1 slider
> pp.slicer(da, dims=['x', 'y'])  # slice out dimensions 'x' and 'y' -> makes a 1d line with 2 sliders
> ```

Comment before having looked at the implementation: I feel this is backwards. Wouldn't it be more natural to specify the dims the user wants to *see*?

SimonHeybrock:
> We are not adding the buttons from the old plotting that allowed to flip the dimensions interactively because this would lead to too much entangled code between the `Mesh` and the `Figure`, where a new `Mesh` would have to be made, the axis limits would have to be changed, and what to do with the `log` scales of the axes.

Would it be an option to make one plot for each axis combination, and show only the selected one? The old buttons next to the slider essentially act like tab-selection buttons, i.e., it simply selects a different figure to show.

nvaytet:
> Comment before having looked at the implementation: I feel this is backwards. Wouldn't it be more natural to specify the dims the user wants to see?

I don't mind either way.
Maybe we should call the argument `keep` like in `sc.collapse`?
```Py
pp.slicer(da, keep=['x'])  # keep only 'x' -> makes a 1d line with 2 sliders
pp.slicer(da, keep=['x', 'y'])  # keep 'x' and 'y' -> makes a 2d image with 1 slider along 'z'
```

nvaytet:
> Would it be an option to make one plot for each axis combination, and show only the selected one? The old buttons next to the slider essentially act like tab-selection buttons, i.e., it simply selects a different figure to show.

I feel this is really adding complication for not much gain. And the different tabs wouldn't help you with the handling of the log scales on the axes.
I really don't think the buttons will be used much, if it's easy for the user to specify when making the plot what they want to keep/slice.

SimonHeybrock:
> > Comment before having looked at the implementation: I feel this is backwards. Wouldn't it be more natural to specify the dims the user wants to see?
> 
> I don't mind either way. Maybe we should call the argument `keep` like in `sc.collapse`?
> 
> ```python
> pp.slicer(da, keep=['x'])  # keep only 'x' -> makes a 1d line with 2 sliders
> pp.slicer(da, keep=['x', 'y'])  # keep 'x' and 'y' -> makes a 2d image with 1 slider along 'z'
> ```

We should probably just get user feedback/opinions.

docs/slicer-plot.ipynb
```diff
@@ -0,0 +1,86 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "a175931f-a0fd-48f7-83cc-9db14d454535",
+   "metadata": {},
+   "source": [
+    "# Slicer plot\n",
```
SimonHeybrock, line 8: As shown by a user comment on Slack, the name "Slicer" may be misleading?

pyproject.toml
```diff
@@ -13,6 +13,7 @@ testpaths = "tests"
 filterwarnings = [
   "error",
   "ignore::UserWarning",
+  "ignore::PendingDeprecationWarning",
```
SimonHeybrock, line 16: Really?

src/plopp/wrappers.py
```diff
@@ -112,3 +120,57 @@ def plot(obj: Union[VariableLike, Dict[str, VariableLike]],
...
+    This will produce one slider per sliced dimension, below the figure.
+
+    Parameters
+    ----------
+    obj:
+        The object to be plotted. Possible inputs are:
+        - Variable
+        - Dataset
+        - DataArray
+        - numpy ndarray
```
SimonHeybrock, line 141: Type hints should take care of this?

pyproject.toml
nvaytet: I don't know how to make this specific to Matplotlib. This arose when updating to Matplotlib 3.6 where `get_cmap` raises this warning. They say to use the syntax
```Py
cmap = matplotlib.colormaps[name]
```
instead of
```Py
cmap = matplotlib.cm.get_cmap(name)
```
However, the new syntax does not exist in `mpl<3.5`. I thought I would keep it compatible?
Alternatively I can just use the new syntax and put a lower bound on the MPL requirement?
What do you think?

docs/slicer-plot.ipynb
nvaytet: Any better ideas?

pyproject.toml
SimonHeybrock: Leave it for now.

docs/slicer-plot.ipynb
SimonHeybrock: Ask the users?

nvaytet: Greg says `slicer` is fine in the end.

nvaytet:
Merging so I can more easily move with other PRs

PR #34, opened 2022-09-29
Add superplot that can save lines on 1d slicer plot

### LINKED ISSUE #15: Add superplot for displaying 1d slices of 2d data, with ability to persist lines

nvaytet:
Example:
```Py
%matplotlib widget
import plopp as pp
da = pp.data.dense_data_array(ndim=2)

pp.superplot(da, keep='xx', ls='solid', marker='None')
```
Screenshot at 2022-09-29 22-12-55

Fixes #15


src/plopp/core/utils.py
```diff
@@ -91,3 +91,10 @@ def merge_masks(masks):
     Combine all masks into a single one using the OR operation.
     """
     return reduce(lambda a, b: a | b, masks.values())
+
+
+def coord_to_string(coord):
+    out = value_to_string(coord.values)
```
SimonHeybrock, line 97: What does this do to long coords?

SimonHeybrock (approved):

src/plopp/core/utils.py
nvaytet: It's only designed to take in either one or two values, in the case of bin edges.
But you are right that I should make that more clear, either in the function name or in a docstring

PR #36, opened 2022-10-06
Spatial cuts in scatter 3d

nvaytet:
This adds a control bar at the bottom of the 3d scatter plot to make spatial cuts along the XYZ axes.
Screenshot at 2022-10-06 16-01-21

- Activating a button makes a cut appear and makes the rest of the scatter points transparent.
- Moving a cut is done using a dedicated slider: it moves the red outline instantly, but uses a debounce mechanism to update the cut only once in a while (if the slider has not moved for 0.3s).
- All 3 cuts can be activated at the same time.
- The +/- buttons can be used to increase/decrease the thickness of the cut.
- The vertical slider in the left toolbar can be used to change the opacity of the original point cloud when at least one cut is active.

**Internal refactor:**

The internals have been refactored significantly in this PR.
This arose when considering the relation between the colormapper and the point clouds or meshes on the figure. Basically, if you have two point clouds on one figure, they must use the same colormapper (the 3d cuts are just additional point clouds in the scene).

If they use the same colormapper, we need to consider what happens when an update is triggered, e.g. with a slider.
The point clouds update one after the other. When the first update happens, the limits on the colorbar potentially change (note that the limits on the colorbar only grow, they do not shrink). We then need to update the colors on both point clouds, using the new colorbar range.

Then, the second point cloud updates, and this may also change the colorbar limits once again. Therefore, everytime the colorbar limits change, we need to update all the point clouds in the scene (and the same goes if you have more than one mesh on your mpl axes, although that is more uncommon).

I first made the colormapper into a node in the graph, which would nicely notify children if its limits have changed.
However, this made setting up custom gaphs with 2d or 3d figures more difficult for the user, because they need to be aware of the colormapper node, when it should just be an implementation detail. It would also show up in the output of `show_graph`, thus confusing the users.

Instead, I went with the following:
- the `Figure` contains a `Canvas` (the mpl axes) and a `ColorMapper`.
- The Canvas just displays objects it receives (`Line`, `Mesh`, `PointCloud`).
- The `Figure` is in charge of performing the updates when notifications come from nodes above.
- The `Figure` gets new data, and updates the range on the `ColorMapper`.
- The `ColorMapper` knows all the meshes or point clouds that depend on it (they are its children), and it updates the colors of all the children (apart from the one that triggered the update) if the colorbar limits have changed.
- The `Figure` then and asks the `ColorMapper` for rgba colors, based on the data received in the update.
- The `Figure` then sets the new colors on the `Mesh` or `PointCloud`

The part I am not so happy with is the `ColorMapper` holding a reference to all its children, but this was the best I could think of without making the `ColorMapper` into a true node in the graph.

**Limiting inheritance:**

Another aspect of this PR is I have tried to limit how many levels of inheritance I have in the classes.
This is a mistake that was made in the old scipp plotting, and we should try to avoid it here.
For example, I was starting to have an `InteractiveFigure2d`, that inherited both from the `VBox` (for display in jupyter) and `Figure2d`. `Figure2` inherited from a common `Figure` class which provided the common functionality between 1d and 2d figures. In turn, the `Figure` also inherited from `View`.

This is why we now have the `Canvas` to provide the common functionality between 1d and 2d figures. The Canvas does not inherit from anyone. The `Figure2d` inherits from `BaseFig`, which is a view, but does not inherit from the basic `View` (the basic `View`, which does not do much, is mostly used for unit tests now).
The `InteractiveFig2d` is now just a `VBox`, which is patched (there were issues upon initialization with multiple inheritance) by adding a `Figure2d` object to it. I then overrode the `__getattr__` so that the user can access the attributes of the underlying figure (e.g. `.canvas`) from the top level, the object that is actually returned by the `plot` function call.


jl-wynen:
Clicking the minus buttons repeatedly triggers an exception which shows up in the Jupyter log console:

```.py
ValueError                                Traceback (most recent call last)
File ~/Work/plopp/src/plopp/widgets/cut3d.py:198, in Cut3dTool.decrease_thickness(self, change)
    196 self.thickness = max(self.thickness - self._thickness_step, 0)
    197 self._remove_cut()
--> 198 self._add_cut()

File ~/Work/plopp/src/plopp/widgets/cut3d.py:169, in Cut3dTool._add_cut(self)
    165 self.select_nodes[n.id] = node(lambda da, pos: da[sc.abs(da.meta[
    166     self._dim] - pos) < sc.scalar(delta, unit=self._unit)])(
    167         da=n, pos=self.pos_nodes[n.id])
    168 self.select_nodes[n.id].add_view(self._view)
--> 169 self._view.update(self.select_nodes[n.id].request_data(),
    170                   key=self.select_nodes[n.id].id,
    171                   colormapper=n.id)

File ~/Work/plopp/src/plopp/graphics/scatterfig.py:87, in ScatterFig.update(self, new_values, key, colormapper)
     85 if colormapper is not None:
     86     colormapper = self._children[colormapper].color_mapper
---> 87 pts = PointCloud(data=new_values,
     88                  x=self._x,
     89                  y=self._y,
     90                  z=self._z,
     91                  colormapper=colormapper,
     92                  **self._kwargs)
     93 self._children[key] = pts
     94 self.scene.add(pts.points)

File ~/Work/plopp/src/plopp/graphics/point_cloud.py:73, in PointCloud.__init__(self, x, y, z, data, colormapper, pixel_size, cmap, mask_cmap, norm, vmin, vmax, opacity)
     67 self.material = p3.PointsMaterial(vertexColors='VertexColors',
     68                                   size=2.5 * pixel_size * pixel_ratio,
     69                                   transparent=True,
     70                                   opacity=opacity)
     71 self.points = p3.Points(geometry=self.geometry, material=self.material)
---> 73 self.color_mapper.set_norm(data=self._data)

File ~/Work/plopp/src/plopp/graphics/color_mapper.py:125, in ColorMapper.set_norm(self, data)
    123 self.unit = data.unit
    124 self.name = data.name
--> 125 self.autoscale(data=data.data, scale='linear')
    126 self.autoscale(data=data.data, scale='log')
    127 self.notify()

File ~/Work/plopp/src/plopp/graphics/color_mapper.py:87, in ColorMapper.autoscale(self, data, scale)
     83 def autoscale(self, data, scale):
     84     """
     85     Re-compute the min and max range of values, given new values.
     86     """
---> 87     vmin, vmax = fix_empty_range(find_limits(data, scale=scale))
     88     if self.user_vmin is not None:
     89         assert self.user_vmin.unit == data.unit

File ~/Work/plopp/src/plopp/core/limits.py:18, in find_limits(x, scale)
     16     finite_min = np.amin(finite_vals, initial=np.inf, where=finite_vals > 0)
     17 else:
---> 18     finite_min = np.amin(finite_vals)
     19 finite_max = np.amax(finite_vals)
     20 return (scalar(finite_min, unit=x.unit), scalar(finite_max, unit=x.unit))

File <__array_function__ internals>:180, in amin(*args, **kwargs)

File ~/.local/share/virtualenvs/plopp-IDxrCsYV/lib/python3.10/site-packages/numpy/core/fromnumeric.py:2918, in amin(a, axis, out, keepdims, initial, where)
   2802 @array_function_dispatch(_amin_dispatcher)
   2803 def amin(a, axis=None, out=None, keepdims=np._NoValue, initial=np._NoValue,
   2804          where=np._NoValue):
   2805     """
   2806     Return the minimum of an array or minimum along an axis.
   2807 
   (...)
   2916     6
   2917     """
-> 2918     return _wrapreduction(a, np.minimum, 'min', axis, None, out,
   2919                           keepdims=keepdims, initial=initial, where=where)

File ~/.local/share/virtualenvs/plopp-IDxrCsYV/lib/python3.10/site-packages/numpy/core/fromnumeric.py:86, in _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs)
     83         else:
     84             return reduction(axis=axis, out=out, **passkwargs)
---> 86 return ufunc.reduce(obj, axis, dtype, out, **passkwargs)

ValueError: zero-size array to reduction operation minimum which has no identity

09.42.07

KeyError                                  Traceback (most recent call last)
File ~/Work/plopp/src/plopp/widgets/cut3d.py:197, in Cut3dTool.decrease_thickness(self, change)
    195 def decrease_thickness(self, change):
    196     self.thickness = max(self.thickness - self._thickness_step, 0)
--> 197     self._remove_cut()
    198     self._add_cut()

File ~/Work/plopp/src/plopp/widgets/cut3d.py:175, in Cut3dTool._remove_cut(self)
    173 def _remove_cut(self):
    174     for key in self.pos_nodes:
--> 175         self._view.remove(self.select_nodes[key].id)
    176         self.select_nodes[key].remove()
    177         self.pos_nodes[key].remove()

File ~/Work/plopp/src/plopp/graphics/scatterfig.py:159, in ScatterFig.remove(self, key)
    155 def remove(self, key):
    156     """
    157     Remove a point cloud from the scene.
    158     """
--> 159     self.scene.remove(self._children[key].points)
    160     del self._children[key]

KeyError: '8aae8cec-4613-11ed-8b23-8cc6813f0665'
```

After that, the plus button does not work anymore either.

nvaytet:
> Clicking the minus buttons repeatedly triggers an exception which shows up in the Jupyter log console:

Ah it's when the thickness is small enough that no points at selected. Good catch!

src/plopp/graphics/color_mapper.py
```diff
@@ -40,52 +45,111 @@ def __init__(self,
...
+                       value=self._norm == 'log',
+                       description='log',
+                       tooltip='Toggle data norm').widget
+        }
+        self._update_colorbar()
+        return ipw.VBox(list(self.colorbar.values()))
+
+    def _update_colorbar(self):
+        dpi = 96
+        height_px = 600
```
jl-wynen, line 74: Should this be tied to the actual height of the axes that it is attached to?

src/plopp/graphics/color_mapper.py
```diff
@@ -40,52 +45,111 @@ def __init__(self,
...
-        elif vmax.value > self.vmax:
-            self.vmax = vmax.value
+            self._vmax[scale] = self.user_vmax.value
+        elif vmax.value > self._vmax[scale]:
+            self._vmax[scale] = vmax.value
 
-        self.norm_func.vmin = self.vmin
-        self.norm_func.vmax = self.vmax
+        self.normalizer[scale].vmin = self._vmin[scale]
+        self.normalizer[scale].vmax = self._vmax[scale]
```
jl-wynen, line 100: This only scales for the given `scale`. When the plot is, e.g. in linear mode and it is auto scaled, and the user then switches to log mode, does it use the scale that was previously used with log? Is this what we want?

src/plopp/graphics/scatterfig.py
```diff
@@ -0,0 +1,160 @@
...
+                                                      'height': '150px'
+                                                  }
+                                              })
+        self.opacity_slider.observe(self._update_opacity, names='value')
+        self.left_bar.add(self.opacity_slider)
+
+        self._original_children = list(self._children.keys())
+        self.cut_x.button.observe(self._toggle_opacity, names='value')
+        self.cut_y.button.observe(self._toggle_opacity, names='value')
+        self.cut_z.button.observe(self._toggle_opacity, names='value')
```
jl-wynen, line 75: Does `ScatterFig` have to contain all these widgets? Or could it be as simple as possible and only contain the actual scatter plot. And there would be a function that creates the `scene` and sets up all widgets as done here?

It might have to connect widgets as graph nodes to set up communication between them.

src/plopp/widgets/cut3d.py
```diff
@@ -0,0 +1,198 @@
...
+                 *nodes,
+                 limits,
+                 direction,
+                 view,
+                 value: bool = False,
+                 color='red',
+                 linewidth=1.5,
+                 **kwargs):
+        """
+        """
```
jl-wynen, line 75: Missing docstring

src/plopp/widgets/cut3d.py
```diff
@@ -0,0 +1,198 @@
...
+        self._direction = direction
+        axis = 'xyz'.index(self._direction)
+        self._dim = self._limits[axis].dim
+        self._unit = self._limits[axis].unit
+        self._view = view
+
+        w_axis = 2 if self._direction == 'x' else 0
+        h_axis = 2 if self._direction == 'y' else 1
+        width = (self._limits[w_axis][1] - self._limits[w_axis][0]).value
+        height = (self._limits[h_axis][1] - self._limits[h_axis][0]).value
```
jl-wynen, line 87: I don't understand what these parameters are. What do 'x' and 'y' mean? And why is there no 'z'?

nvaytet: So in order to make a frame/outline, we have to first make the outline (which by default has the z direction as the vector normal to the plane), and then we rotate it. I have used the following scheme:

x as normal vector:
- width: dz (axis=2)
- height: dy (axis=1)
- rotate around Y axis

y as normal vector:
- width: dx (axis=0)
- height: dz (axis=2)
- rotate around X axis

z as normal vector:
- width: dx (axis=0)
- height: dy (axis=1)
- no rotate

src/plopp/graphics/color_mapper.py
nvaytet: Because the `ColorMapper` is not directly aware of the figure it is used in, it was quite messy to pass around the figure size everywhere. So instead I picked an average size, and then I resize it here: https://github.com/scipp/plopp/blob/slicing_in_3d/src/plopp/graphics/scatterfig.py#L106
(with some scaling factor to allow some space for the `log` button)

It would look cleaner if we used the real size of the figure directly, as it doesn't always look super when it scales the image down or up, but I went for code simplicity instead of looks. We can change.

src/plopp/functions/scatter3d.py
```diff
@@ -13,7 +13,7 @@ def scatter3d(da: sc.DataArray,
               y: str = None,
               z: str = None,
               pos: str = None,
-              figsize: Tuple[Union[int, float]] = None,
+              figsize: Tuple[Union[int, float]] = (600, 400),
```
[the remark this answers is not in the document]
nvaytet: I changed it back to be `None` here.

src/plopp/graphics/color_mapper.py
nvaytet: This was my way of not needing to request the data again when toggling the log. You compute the limits for both linear and log when you first make the figure, and store those. But it is true that it will use old limits if you then move a slider and change the colors, and then switch to log, your limits will be out of date (I think?)... I should probably change.

src/plopp/graphics/color_mapper.py
jl-wynen: It is fine for now. We can do this kind of polishing later when we have a more or less finished layout.

src/plopp/graphics/fig2d.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        if new_values.ndim != 2:
+            raise ValueError("Figure2d can only be used to plot 2-D data.")
+
+        self.colormapper.update(data=new_values, key=key)
+
+        if key not in self._children:
+
+            mesh = Mesh(canvas=self.canvas, data=new_values)
+            self._children[key] = mesh
```
nvaytet, line 57: rename `_children` to artists?

src/plopp/graphics/mesh.py
```diff
@@ -106,77 +99,33 @@ def __init__(self,
...
+        return out
+
+    def set_colors(self, rgba):
         self._mesh.set_facecolors(rgba.reshape(np.prod(rgba.shape[:-1]), 4))
 
     def update(self, new_values: sc.DataArray):
         """
         Update image array with new values.
         """
         self._data = new_values
```
nvaytet, line 131: Is update on mesh still needed?

src/plopp/graphics/interactive.py
```diff
@@ -39,65 +41,126 @@ def _is_sphinx_build():
...
 
-class InteractiveFig(Figure, VBox):
+class InteractiveFig1d(VBox):
 
     def __init__(self, *args, **kwargs):
 
-        Figure.__init__(self, *args, **kwargs)
-        VBox.__init__(self, [
+        self._fig = Figure1d(*args, **kwargs)
+        self._fig.canvas.fig.canvas.toolbar_visible = False
```
nvaytet, line 49: have a function on `Canvas` that does this internally

src/plopp/graphics/mesh.py
nvaytet: Yes, this needs to remain because we get the `.data` from the mesh to send to the `rgba`, and that `.data` could be different from the `new_values` in the case of a 2D coord.

src/plopp/graphics/colormapper.py
```diff
@@ -114,11 +114,18 @@ def autoscale(self, data):
...
-        for child in self.children.values():
-            child.set_colors(self.rgba(child.data))
+    def _set_children_colors(self, key=None):
+        """
+        Update the colors of all the children apart from the one that triggered the
+        update, as those get updated by the figure.
+        """
+        keys = set(self.children.keys())
+        if key is not None:
+            keys -= set([key])
```
jl-wynen, line 124:
```suggestion
        keys = set(self.children.keys()) - {key}
```
no need to check for `None` unless that is a valid key (?)

jl-wynen (approved):
Just some minor things. Looks good overall.

We need a lot more docstrings, though. But that can be done later.

src/plopp/graphics/interactive.py
```diff
@@ -75,6 +75,9 @@ def __getattr__(self, key):
         except AttributeError:
             return getattr(self._fig, key)
 
+    def __dir__(self):
+        return list(set(dir(VBox) + dir(super()) + dir(self._fig)))
```
jl-wynen, line 79: Technically also `self.left_bar`, etc. Are those actually part of the public API?

nvaytet: Those are listed by the `dir(super())`, strangely enough...

nvaytet: I added a `sorted` now.

PR #52, opened 2022-10-21
Add support for string coords

### LINKED ISSUE #40: Exception when plotting data with string coordinate
```python
TypeError                                 Traceback (most recent call last)
Cell In [4], line 1
----> 1 table.plot()

File ~/code/scipp/plopp/src/plopp/functions/plot.py:94, in plot(obj, aspect, cbar, crop, errorbars, grid, ignore_size, mask_color, norm, scale, title, vmin, vmax, **kwargs)
     92     return figure(*nodes, **all_args)
     93 else:
---> 94     return figure(input_node(preprocess(obj, crop=crop, ignore_size=ignore_size)),
     95                   **all_args)

File ~/code/scipp/plopp/src/plopp/functions/figure.py:16, in figure(*args, **kwargs)
     14 else:
     15     from ..graphics import StaticFig
---> 16     return StaticFig(*args, **kwargs)

File ~/code/scipp/plopp/src/plopp/graphics/fig.py:65, in Figure.__init__(self, ax, figsize, title, grid, vmin, vmax, norm, aspect, scale, cbar, errorbars, mask_color, crop, *nodes, **kwargs)
     62 self._new_artist = False
     64 self._post_init()
---> 65 self.render()

File ~/code/scipp/plopp/src/plopp/graphics/fig.py:222, in Figure.render(self)
    220     new_values = node.request_data()
    221     self.update(new_values=new_values, key=node.id, draw=False)
--> 222 self._autoscale()
    223 self.crop(**self._crop)
    224 self.draw()

File ~/code/scipp/plopp/src/plopp/graphics/fig.py:87, in Figure._autoscale(self)
     85 yscale = self._ax.get_yscale()
     86 for child in self._children.values():
---> 87     xlims, ylims = child.get_limits(xscale=xscale, yscale=yscale)
     88     if isinstance(child, Line):
     89         if self._user_vmin is not None:

File ~/code/scipp/plopp/src/plopp/graphics/line.py:182, in Line.get_limits(self, xscale, yscale)
    178 def get_limits(self, xscale, yscale) -> Tuple[float, ...]:
    179     """
    180     """
    181     xmin, xmax = fix_empty_range(
--> 182         find_limits(self._data.meta[self._dim], scale=xscale))
    183     ymin, ymax = fix_empty_range(find_limits(self._data.data, scale=yscale))
    185     for lim in (xmin, xmax, ymin, ymax):

File ~/code/scipp/plopp/src/plopp/core/limits.py:14, in find_limits(x, scale)
     10 """
     11 Find sensible limits, depending on linear or log scale.
     12 """
     13 v = x.values
---> 14 finite_vals = v[np.isfinite(v)]
     15 if scale == "log":
     16     finite_min = np.amin(finite_vals, initial=np.inf, where=finite_vals > 0)

TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

nvaytet:
1D plots were mostly working after #36 , but some fixes were needed for 2d plots.

When string coord is bin centers:
```Py
strings = ['a', 'b', 'c', 'd', 'e']
da = sc.DataArray(data=sc.array(dims=['y', 'x'], values=np.random.random((5, 5))),
                  coords={
                      'x': sc.array(dims=['x'], values=strings, unit='s'),
                      'y': sc.arange('y', 6., unit='m')
                  })
pp.plot(da)
```
Screenshot at 2022-10-23 20-59-26

When string coord is bin edges:
```Py
strings = ['a', 'b', 'c', 'd', 'e', 'f']
da = sc.DataArray(data=sc.array(dims=['y', 'x'], values=np.random.random((5, 5))),
                  coords={
                      'x': sc.array(dims=['x'], values=strings, unit='s'),
                      'y': sc.arange('y', 6., unit='m')
                  })
pp.plot(da)
```
Screenshot at 2022-10-23 21-00-09

Fixes #40

### DIFF
--- a/src/plopp/graphics/line.py
+++ b/src/plopp/graphics/line.py
@@ -125,9 +125,8 @@ class Line:
             if data["mask"] is not None:
                 data["mask"]["y"] = np.concatenate(
                     (data["mask"]["y"][0:1], data["mask"]["y"]))
-            data["variances"]["x"] = 0.5 * (x[1:] + x[:-1])
-        else:
-            data["variances"]["x"] = x
+        if self._data.variances is not None:
+            data["variances"]["x"] = 0.5 * (x[1:] + x[:-1]) if hist else x
         data["variances"]["y"] = y
         data["hist"] = hist
         return data
--- a/src/plopp/core/utils.py
+++ b/src/plopp/core/utils.py
@@ -18,7 +18,6 @@ def coord_as_bin_edges(da, key, dim=None):
         x = sc.arange(dim, float(x.shape[0]), unit=x.unit)
     if da.meta.is_edges(key, dim=dim):
         return x
-    # idim = x.dims.index(dim)
     if x.sizes[dim] < 2:
         half = sc.scalar(0.5, unit=x.unit)
         return sc.concat([x[dim, 0:1] - half, x[dim, 0:1] + half], dim)
--- a/src/plopp/core/utils.py
+++ b/src/plopp/core/utils.py
@@ -50,7 +50,7 @@ def number_to_variable(x):
     """
     Convert the input int or float to a variable.
     """
-    return scalar(x, unit=None) if isinstance(x, (int, float)) else x
+    return sc.scalar(x, unit=None) if isinstance(x, (int, float)) else x
 
 
 def name_with_unit(var=None, name=None, log=False):
--- a/src/plopp/core/utils.py
+++ b/src/plopp/core/utils.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
 
-from scipp import scalar, concat, midpoints
+import scipp as sc
 import uuid
 from functools import reduce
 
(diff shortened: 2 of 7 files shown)


src/plopp/graphics/line.py
```diff
@@ -125,9 +125,8 @@ def _preprocess_hist(self, data: dict) -> Tuple[dict, bool]:
             if data["mask"] is not None:
                 data["mask"]["y"] = np.concatenate(
                     (data["mask"]["y"][0:1], data["mask"]["y"]))
-            data["variances"]["x"] = 0.5 * (x[1:] + x[:-1])
-        else:
-            data["variances"]["x"] = x
+        if self._data.variances is not None:
+            data["variances"]["x"] = 0.5 * (x[1:] + x[:-1]) if hist else x
```
SimonHeybrock, line 129: Um, what is this doing?

SimonHeybrock (commented):
Looks good, but one question:

src/plopp/graphics/line.py
nvaytet: It's placing the error bar in the middle of the bin instead of on the edges, when the coord is bin edges and the line is a step function.

nvaytet:
One thing I forgot to mention.
Unlike in Scipp, when the coord with strings is long, it will place a tick for every entry.
This is also what Matplotlib does.

If this is a no-go, when we will have to think about using axis formatters again, as in Scipp.
Screenshot at 2022-10-24 07-48-46

src/plopp/graphics/line.py
SimonHeybrock: Ok, so the "x" is the "x" position of the error bars, not the "x" coordinate?

nvaytet: yes

SimonHeybrock (approved):

PR #54, opened 2022-10-24
Fix kwargs that were not being forwarded to mesh

nvaytet:
With this, you can now do, e.g.
```Py
da = pp.data.dense_data_array(ndim=2)
pp.plot(da, rasterized=False)
```


SimonHeybrock (commented):
Test would be useful

SimonHeybrock (approved):

PR #55, opened 2022-10-25
Add dpi option to savefig

nvaytet:
Usage:
```Py
p = pp.plot(da)
p.canvas.savefig(filename, dpi=600)
```

### DIFF
--- a/src/plopp/graphics/canvas.py
+++ b/src/plopp/graphics/canvas.py
@@ -115,14 +115,14 @@ class Canvas:
     def draw(self):
         self.fig.canvas.draw_idle()
 
-    def savefig(self, filename: str, dpi: float = None):
+    def savefig(self, filename: str, **kwargs):
         """
         Save plot to file.
         Possible file extensions are `.jpg`, `.png` and `.pdf`.
         The default directory for writing the file is the same as the
         directory where the script or notebook is running.
         """
-        self.fig.savefig(filename, bbox_inches="tight", dpi=dpi)
+        self.fig.savefig(filename, **{**{'bbox_inches': 'tight'}, **kwargs})
 
     def show(self):
         self.fig.show()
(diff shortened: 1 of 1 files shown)


src/plopp/graphics/canvas.py
```diff
@@ -115,14 +115,14 @@ def autoscale(self, draw=True):
     def draw(self):
         self.fig.canvas.draw_idle()
 
-    def savefig(self, filename: str = None):
+    def savefig(self, filename: str, dpi: float = None):
```
SimonHeybrock, line 118: Any reason we don't just forward all kwargs?

nvaytet: No reason, i'll change

src/plopp/graphics/canvas.py
```diff
@@ -115,14 +115,14 @@ def autoscale(self, draw=True):
...
-    def savefig(self, filename: str, dpi: float = None):
+    def savefig(self, filename: str, **kwargs):
         """
         Save plot to file.
         Possible file extensions are `.jpg`, `.png` and `.pdf`.
         The default directory for writing the file is the same as the
         directory where the script or notebook is running.
         """
-        self.fig.savefig(filename, bbox_inches="tight", dpi=dpi)
+        self.fig.savefig(filename, **{**{'bbox_inches': 'tight'}, **kwargs})
```
SimonHeybrock, line 125:
```suggestion
        settings = {'bbox_inches': 'tight'}
        settings.update(kwargs)
        self.fig.savefig(filename, **settings})
```
or something like that, so users can override our defaults?

nvaytet: users can override the default with the syntax I used. If `bbox_inches` is in kwargs, it will take precedence over our default

SimonHeybrock: Oh, I didn't notice the nested '**{**'{...}'

SimonHeybrock (approved):

PR #57, opened 2022-10-26
Configure dependabot for auto-updating scipp

SimonHeybrock:


.github/dependabot.yml
```diff
@@ -0,0 +1,17 @@
...
+    schedule:
+      interval: "weekly"
+    ignore:
+      # Optional: Official actions have moving tags like v1;
+      # if you use those, you don't need updates.
+      - dependency-name: "actions/*"
+  - package-ecosystem: "pip"
+    directory: "/requirements"
+    schedule:
+      interval: "daily"
```
nvaytet, line 14: is daily necessary, or would weekly be enough?

SimonHeybrock: If a Scipp update breaks Plopp, we want to know quickly as it has impact on users.

nvaytet (approved):

PR #58, opened 2022-10-27
Make figure images not blurry in docs

nvaytet:
Left: after, Right: before
Screenshot at 2022-10-27 10-06-48

### DIFF
--- a/src/plopp/graphics/canvas.py
+++ b/src/plopp/graphics/canvas.py
@@ -71,14 +71,7 @@ class Canvas:
 
     def to_image(self):
         from ipywidgets import Image
-        bounds = self.fig.get_tightbbox(self.fig.canvas.get_renderer()).bounds
-        width = bounds[2] - bounds[0]
-        height = bounds[3] - bounds[1]
-        dpi = self.fig.get_dpi()
-        return Image(value=fig_to_bytes(self.fig),
-                     width=width * dpi,
-                     height=height * dpi,
-                     format='png')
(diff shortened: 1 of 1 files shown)


src/plopp/graphics/canvas.py
```diff
@@ -71,14 +71,7 @@ def __init__(self,
...
         from ipywidgets import Image
-        bounds = self.fig.get_tightbbox(self.fig.canvas.get_renderer()).bounds
-        width = bounds[2] - bounds[0]
-        height = bounds[3] - bounds[1]
-        dpi = self.fig.get_dpi()
-        return Image(value=fig_to_bytes(self.fig),
-                     width=width * dpi,
-                     height=height * dpi,
-                     format='png')
+        return Image(value=fig_to_bytes(self.fig), format='png')
```
SimonHeybrock, line 74: Why is this affecting interactive plots with widgets?

nvaytet: It isn't. Those are screenshots from the docs. By default, we just have svg figures in the docs.
However, when a notebook requires the `%matplotlib widget` backend to work (if a user downloads the notebook for example), the interactive figure won't render alongside the other widgets in sphinx.

We therefore have to convert it to a widget image, and place that in the widget containers. See [here](https://github.com/scipp/plopp/blob/main/src/plopp/graphics/interactive.py#L36) for example.

SimonHeybrock (approved):

PR #64, opened 2022-10-31
Fix ax arguments

### LINKED ISSUE #63: ax arg in 2d plots is broken
```python
fig, ax = plt.subplots(1, 1)
pp.plot(sc.DataArray(sc.zeros(sizes={'x': 10, 'y': 10})), ax=ax)
```
raises
```

File ~/Software/mambaforge/envs/uncertain/lib/python3.8/site-packages/plopp/graphics/mesh.py:100, in Mesh.__init__(self, canvas, data, shading, rasterized, **kwargs)
     94 self._mesh = None
     96 x, y, z = _from_data_array_to_pcolormesh(data=self._data.data,
     97                                          coords=bin_edge_coords,
     98                                          dim_1d=self._dim_1d,
     99                                          dim_2d=self._dim_2d)
--> 100 self._mesh = self._ax.pcolormesh(x.values,
    101                                  y.values,
    102                                  z.values,
    103                                  shading=shading,
    104                                  rasterized=rasterized,
    105                                  **kwargs)
    106 self._mesh.set_array(None)
    108 for xy, var in string_labels.items():

TypeError: inner() got multiple values for argument 'ax'
```

nvaytet:
Fix #63 

Note that I have used `Any` for the axes types in `Figure1d` and `Figure2d`, to hide any Matplotlib details from the figures.
Could change to `matplotlib.pyplot.Axes` if we think that is better.


jl-wynen:
The spacing seems to be wrong:
ax-arg

Note that I need to use
```python
%matplotlib widget
plt.ioff()
```
because otherwise, I get 2 plots. This is not the case without the `ax` argument.

nvaytet:
> Note that I need to use
%matplotlib widget
plt.ioff()
because otherwise, I get 2 plots. This is not the case without the ax argument.

Yes, that is expected because matplotlib does its own thing. So as soon as you call `plt.subplots()`, it will show you the axes after the code cell. Then, because the last thing in the cell is the call to `plot`, it will also display the repr of the figure.
You can also do
```Py
fig, ax = plt.subplots()
p = pp.plot(..., ax=ax)
```
and you will only have one figure.

nvaytet:
> The spacing seems to be wrong:

I'm not too sure what to do in this particular case. If someone has carefully made some axes, and then sends them to me, I don't really want to modify them (meaning I don't want to resize them).

Matplotlib's default behaviour is that if you give it some axes, and then ask for a colorbar, it will reduce the width of the axes you supplied to make space for the colorbar. What I did here was to just add the colorbar to the right, without modifying the input axes. I don't really know what is better. My choice works ok with static figures, but yes, it clips the tick labels on the colorbar with the widget backend.

Screenshot at 2022-11-01 15-06-08

If I then save the figure from the widget backend to file, the figure looks fine, no clipping. So the issue is how `ipympl` determines the bounding box of the figure.

nvaytet:
I am now resizing existing axes and also added an argument to control the width of the colorbar.
Screenshot at 2022-11-01 15-42-04

jl-wynen:
Maybe tangential, but `tight_layout` does not work. Note the warning.

tight

nvaytet:
@jl-wynen I changed strategy for making the colorbar axes. Can you have another look?

jl-wynen (approved):
Works in my tests, including `tight_layout`. And it even looks simpler than the old solution :+1:

PR #68, opened 2022-11-01
Fix vmin vmax in 1d figures

### LINKED ISSUE #62: vmin and vmax not working on 1d plots

nvaytet:
Fix #62 

Screenshot at 2022-11-01 17-18-20

### DIFF
--- a/src/plopp/graphics/canvas.py
+++ b/src/plopp/graphics/canvas.py
@@ -138,8 +138,6 @@ class Canvas:
                 self._xmax = max(self._xmax, right.value)
                 self._ymin = min(self._ymin, bottom.value)
                 self._ymax = max(self._ymax, top.value)
-        # TODO: if searching for limits becomes a performance issue, user specified
-        # y-axis limits via vmin/vmax could be taken into account to speed things up.
         if self._user_vmin is not None:
             assert self._user_vmin.unit == self.yunit
             self._ymin = self._user_vmin.value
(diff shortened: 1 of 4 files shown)


src/plopp/graphics/canvas.py
```diff
@@ -136,6 +138,14 @@ def autoscale(self, draw: bool = True):
                 self._xmax = max(self._xmax, right.value)
                 self._ymin = min(self._ymin, bottom.value)
                 self._ymax = max(self._ymax, top.value)
+        # TODO: if searching for limits becomes a performance issue, user specified
+        # y-axis limits via vmin/vmax could be taken into account to speed things up.
```
SimonHeybrock, line 142: Where would you expect a performance issue?

SimonHeybrock (approved):

src/plopp/graphics/canvas.py
nvaytet: My thinking is that in the code above those lines, we do a search on all the lines and collections in the plot for the bounds.
If there are many lines, or if the lines have many points or the collections many patches, this search could get quite slow (this is independent of the changes in this PR).

If the user specifies vmin and vmax for example, this search could just be skipped because we overwrite the values anyway. I'll make the comment clearer.

nvaytet: Although that only works for the y-axis limits, and we still need to autoscale for the x-axis...

nvaytet: I'll just remove the comment I think.

PR #79, opened 2022-11-08
Fix conversion to bin edges for ints

nvaytet:


src/plopp/core/utils.py
```diff
@@ -29,6 +29,8 @@ def coord_as_bin_edges(da: sc.DataArray, key: str, dim: str = None) -> sc.Variab
         x = sc.arange(dim, float(x.shape[0]), unit=x.unit)
     if da.meta.is_edges(key, dim=dim):
         return x
+    if x.dtype == int:
+        x = x.to(dtype='float64')
```
SimonHeybrock, line 33: It would be nice to keep axis tic labels as integers if the input is, does this influence this?

nvaytet: Matplotlib convertes everything to float and also adds ticks in-between the integers if it decides it needs more ticks.
I don't think it's worth trying to fix.
Screenshot at 2022-11-08 10-17-15

src/plopp/core/utils.py
```diff
@@ -29,6 +29,8 @@ def coord_as_bin_edges(da: sc.DataArray, key: str, dim: str = None) -> sc.Variab
         x = sc.arange(dim, float(x.shape[0]), unit=x.unit)
     if da.meta.is_edges(key, dim=dim):
         return x
+    if x.dtype == int:
```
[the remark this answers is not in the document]
nvaytet: Just checked, it doesn't work for int32. Nice catch

SimonHeybrock (approved):

PR #80, opened 2022-11-08
Switch to PNG repr for static figure when there are too many markers in plot

### LINKED ISSUE #78: Data size limits and SVG output
Consider
```python
import scipp as sc
import plopp

data = sc.ones(dims=['x'], shape=[240_000])
sc.plot(data)
plopp.plot(data)
```
Scipp plots reasonably quickly, but the Plopp plot takes much longer to show, and even creates lag in the notebook after creation. I think the plotting itself is not the culprit, but the actual display in the notebook.

We could reduce the limit until which we can plot, but since it may "just" be a problem with the SVG output, we may consider switching to PNG for large data, or at least provide an option/recommendation?

nvaytet:
Use a switch to a PNG repr if there are too many points or lines drawn on the figure, as rendering the full SVG slows down the notebook.
One can forcibly request a format by using
```Py
pp.plot(da, form='svg')
```

Fix #78

### DIFF
--- a/src/plopp/graphics/fig1d.py
+++ b/src/plopp/graphics/fig1d.py
@@ -48,6 +48,10 @@ class Figure1d(BaseFig):
     ax:
         If supplied, use these axes to create the figure. If none are supplied, the
         canvas will create its own axes.
+    form:
+        Format of the figure displayed in the Jupyter notebook. If ``None``, a SVG is
+        created as long as the number of markers in the figure is not too large. If too
+        many markers are drawn, a PNG image is created instead.
     **kwargs:
         All other kwargs are forwarded to Matplotlib:
 
@@ -69,6 +73,7 @@ class Figure1d(BaseFig):
                  title: str = None,
                  figsize: Tuple[float, float] = (6., 4.),
                  ax: Any = None,
+                 form: Literal['svg', 'png'] = None,
                  **kwargs):
 
         super().__init__(*nodes)
@@ -77,6 +82,7 @@ class Figure1d(BaseFig):
         self._errorbars = errorbars
         self._mask_color = mask_color
         self._kwargs = kwargs
+        self._repr_format = form
         self.canvas = Canvas(cbar=False,
                              aspect=aspect,
                              grid=grid,
--- a/src/plopp/graphics/fig1d.py
+++ b/src/plopp/graphics/fig1d.py
@@ -7,7 +7,7 @@ from .canvas import Canvas
 from .line import Line
 
 import scipp as sc
-from typing import Any, Dict, Literal, Tuple, Union
+from typing import Any, Dict, Literal, Optional, Tuple, Union
 
 
 class Figure1d(BaseFig):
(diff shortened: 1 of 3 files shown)


src/plopp/graphics/fig1d.py
```diff
@@ -48,6 +48,10 @@ class Figure1d(BaseFig):
     ax:
         If supplied, use these axes to create the figure. If none are supplied, the
         canvas will create its own axes.
+    form:
```
SimonHeybrock, line 51:
```suggestion
    format:
```
would be much easier to remember, as it matches the `savefig` param name

src/plopp/graphics/fig1d.py
```diff
@@ -69,6 +73,7 @@ def __init__(self,
                  title: str = None,
                  figsize: Tuple[float, float] = (6., 4.),
                  ax: Any = None,
+                 form: Literal['svg', 'png'] = None,
```
SimonHeybrock, line 76:
```suggestion
                 form: Optional[Literal['svg', 'png']] = None,
```
?

src/plopp/graphics/fig2d.py
```diff
@@ -57,6 +57,10 @@ class Figure2d(BaseFig):
     cax:
         If supplied, use these axes for the colorbar. If none are supplied, and a
         colorbar is required, the canvas will create its own axes.
+    form:
```
SimonHeybrock, line 60:
```suggestion
    format:
```

src/plopp/graphics/fig2d.py
```diff
@@ -77,12 +81,14 @@ def __init__(self,
                  figsize: Tuple[float, float] = (6., 4.),
                  ax: Any = None,
                  cax: Any = None,
+                 form: Literal['svg', 'png'] = None,
```
SimonHeybrock, line 84: optional?

src/plopp/graphics/static.py
```diff
@@ -16,10 +27,16 @@ def _repr_mimebundle_(self, include=None, exclude=None) -> dict:
...
-            'image/svg+xml': fig_to_bytes(self.canvas.fig, form='svg').decode()
-        }
+        out = {'text/plain': 'Figure'}
+        if self._repr_format is not None:
+            repr_maker = REPR_MAP[self._repr_format.lower()]
+        else:
+            npoints = 0
+            for line in self.canvas.ax.lines:
+                npoints += len(line.get_xdata())
+            repr_maker = REPR_MAP['png' if (npoints > 60_000) else 'svg']
```
SimonHeybrock, line 37: Why 60k specifically? Is it fast below that?

SimonHeybrock (changes requested):

src/plopp/graphics/fig1d.py
nvaytet: I avoided format because it is a special name.
Screenshot at 2022-11-08 12-21-59

So doing `self._format = format` is not the best.
Thoughts?

SimonHeybrock: Personally I think it is fine, @jl-wynen?

src/plopp/graphics/static.py
nvaytet: Just trial and error on one machine. I don't have a good justification. Would probably be best to have something like "when the size of the svg output is more than the png bytes", but I don't really know how to predict that without first making both and comparing them, which seems counter-productive...

SimonHeybrock: 60k SVG elements seems a lot. Have you considered, e.g., 10k?

SimonHeybrock (approved):

PR #81, opened 2022-11-08
Fixes for scatter3d plots

nvaytet:
- Factor in pixel size to compute point cloud limits, so that a flat panel of pixels has correct size in all 3 directions
- Flatten input data with ndim > 1 so that 3d spatial cuts can work (selection broke down on multi-d coordinates)

### DIFF
--- a/tests/graphics/point_cloud_test.py
+++ b/tests/graphics/point_cloud_test.py
@@ -27,11 +27,26 @@ def test_update():
 
 def test_get_limits():
     da = scatter_data()
-    cloud = PointCloud(data=da, x='x', y='y', z='z')
+    pix = 0.5
+    cloud = PointCloud(data=da, x='x', y='y', z='z', pixel_size=pix)
     xlims, ylims, zlims = cloud.get_limits()
-    assert sc.identical(xlims[0], da.meta['x'].min())
-    assert sc.identical(xlims[1], da.meta['x'].max())
-    assert sc.identical(ylims[0], da.meta['y'].min())
-    assert sc.identical(ylims[1], da.meta['y'].max())
-    assert sc.identical(zlims[0], da.meta['z'].min())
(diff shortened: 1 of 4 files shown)


tests/graphics/point_cloud_test.py
```diff
@@ -27,11 +27,26 @@ def test_update():
...
-    cloud = PointCloud(data=da, x='x', y='y', z='z')
+    pix = 0.5
+    cloud = PointCloud(data=da, x='x', y='y', z='z', pixel_size=pix)
+    xlims, ylims, zlims = cloud.get_limits()
+    assert sc.identical(xlims[0], da.meta['x'].min() - sc.scalar(0.5 * pix, unit='m'))
+    assert sc.identical(xlims[1], da.meta['x'].max() + sc.scalar(0.5 * pix, unit='m'))
+    assert sc.identical(ylims[0], da.meta['y'].min() - sc.scalar(0.5 * pix, unit='m'))
+    assert sc.identical(ylims[1], da.meta['y'].max() + sc.scalar(0.5 * pix, unit='m'))
+    assert sc.identical(zlims[0], da.meta['z'].min() - sc.scalar(0.5 * pix, unit='m'))
+    assert sc.identical(zlims[1], da.meta['z'].max() + sc.scalar(0.5 * pix, unit='m'))
```
SimonHeybrock, line 38: This is not a very useful test, maybe? This is just checking an implementation detail, and with every tiny code modification you will need to update the test.

SimonHeybrock (approved):

tests/graphics/point_cloud_test.py
nvaytet: I'll see. The `Outline` depends on this working properly, but I agree that the usefulness factor is not through the roof.
The next time it breaks I'll remove it.

PR #83, opened 2022-11-09
Fix extra figure when `cbar=False`

### LINKED ISSUE #77: `cbar=False` opens additional figure for the colorbar
Using `cbar=False` with a regular mpl backend (w/o) Jupyter creates a new figure for the colorbar. E.g. the following opens 2 windows, one that contains the plot w/o colorbar and one containing only the colorbar.
```python
import plopp as pp
pp.plot(pp.data.data_array(ndim=2), cbar=False)
plt.show()
```

nvaytet:
Fix #77

### DIFF
--- a/src/plopp/graphics/colormapper.py
+++ b/src/plopp/graphics/colormapper.py
@@ -56,7 +56,8 @@ class ColorMapper:
         The axes to use for the colorbar. If none are supplied, the ColorMapper will
         create its own axes.
     cbar:
-        Create a colorbar if ``True``.
+        Create a colorbar if ``True``. If ``False``, no colorbar is made even if ``cax``
+        is defined.
     cmap:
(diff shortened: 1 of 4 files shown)


src/plopp/graphics/colormapper.py
```diff
@@ -55,6 +55,8 @@ class ColorMapper:
     cax:
         The axes to use for the colorbar. If none are supplied, the ColorMapper will
         create its own axes.
+    cbar:
+        Create a colorbar if ``True``.
```
jl-wynen, line 59: This argument has no effect if `cax` is set. Is this intentional? If so, please document.

nvaytet: Not intentional, i'll fix.

jl-wynen (approved):

PR #84, opened 2022-11-11
Raise error in limits when all is nan

nvaytet:


tests/core/limits_test.py
```diff
@@ -44,11 +45,12 @@ def test_find_limits_with_ninf():
...
-def test_find_limits_no_finite_values():
+def test_find_limits_no_finite_values_raises():
     x = sc.array(dims=['x'], values=[np.nan, np.inf, np.NINF, np.nan], unit='m')
-    lims = find_limits(x)
-    assert sc.isnan(lims[0])
-    assert sc.isnan(lims[1])
+    with pytest.raises(ValueError) as e:
+        _ = find_limits(x)
+    assert str(e.value) == ("No finite values were found in array. "
+                            "Cannot compute limits.")
```
jl-wynen, line 53: You can use the `match` argument of `pytest.raises` instead.

jl-wynen (approved):
Only a minor thing, do it or leave it

tests/core/limits_test.py
nvaytet: Didn't know that arg, thanks. I applied it in all tests.

PR #85, opened 2022-11-11
Refactor inspector plot to use an InspectTool

nvaytet:
We refactor the inspector plot to use an `InspectTool` to make it easier in the future for using alongside other tools.


src/plopp/widgets/linecut_tool.py
```diff
@@ -0,0 +1,121 @@
...
+            if len(figure.graph_nodes) > 1:
+                raise ValueError("Cannot automatically generate tool because figure "
+                                 "has more than one artist.")
+            data = list(figure.graph_nodes.values())[0].request_data()
+        self._data_array = data
+        if (self._data_array.coords[figure.dims['x']].unit == self._data_array.coords[
+                figure.dims['y']].unit):
+            distance_unit = self._data_array.coords[figure.dims['x']].unit
+        else:
+            distance_unit = ''
```
SimonHeybrock, line 80: This is a bit weird. Maybe a unit mismatch would indicate that doing this is just conceptually meaningless anyway?

src/plopp/widgets/linecut_tool.py
```diff
@@ -0,0 +1,121 @@
...
+def interpolator(xy, interp_func, resolution, data_unit, distance_unit):
+    x1 = xy['x'][0]
+    y1 = xy['y'][0]
+    x2 = xy['x'][1]
+    y2 = xy['y'][1]
+
+    xvalues = np.linspace(x1, x2, resolution)
+    yvalues = np.linspace(y1, y2, resolution)
+
+    profile = np.array([interp_func(x, y)[0] for x, y in zip(xvalues, yvalues)])
```
SimonHeybrock, line 25: Why are you doing this in a loop?

src/plopp/widgets/linecut_tool.py
```diff
@@ -0,0 +1,121 @@
...
+    out = sc.DataArray(data=sc.array(dims=['distance'], values=profile, unit=data_unit),
+                       coords={
+                           'distance':
+                           sc.array(dims=['distance'],
+                                    values=np.linalg.norm(
+                                        [xvalues - xvalues[0], yvalues - yvalues[0]],
+                                        axis=0),
+                                    unit=distance_unit)
```
SimonHeybrock, line 33: I don't think this a useful coordinate in practice? I can imagine someone might use this on an image, or a I(Qx, Qy, Qz) intensity, but in either case the user would want to know the coordinates.

SimonHeybrock (commented):
Not sure how useful this is for us in practice. The applications I am aware of would typically need to integrate/sum along a user-defined width perpendicular to the selected line. In many applications we would also need axis-alignment, i.e., lines would either be horizontal or vertical.

nvaytet:
I think you're right that it's not very useful and a little I'll defined (the part with the units). I'll remove the Linecut tool (maybe move it to a specific example in the gallery where units make sense?) and keep the refactor around the Inspect tool

nvaytet:
I'm having further doubts about this, and I have new ideas on how to make a generic interface to `mpltoolbox`. I also think I am slowly drifting away from one of the things I saw we would do to start with, which is making single purpose visualizations, rather than things that can work with everything else.

So I am thinking that the inspector plot should maybe remain self-contained?

Marking as draft for now.

nvaytet:
Closing in favour of #103

PR #86, opened 2022-11-17
Support datasets and dict of data arrays in slicer

### LINKED ISSUE #61: Slicer plot does not work with dict of data arrays or dataset
```Py
import plopp as pp
import scipp as sc

da = pp.data.dataset(ndim=2)
pp.slicer(da, keep=['xx'])
```
fails with `AttributeError: 'scipp._scipp.core.Dataset' object has no attribute 'name'`

nvaytet:
Fix #61

### DIFF
--- a/src/plopp/functions/superplot.py
+++ b/src/plopp/functions/superplot.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
 
 from .common import require_interactive_backend, preprocess
-# from .slicer import Slicer
+from .slicer import Slicer
 from ..core import Node, View
 from ..core.utils import coord_element_to_string
 
@@ -103,9 +103,9 @@ def superplot(obj: Union[sc.typing.VariableLike, ndarray],
     da = preprocess(obj, crop=crop, ignore_size=True)
     if keep is None:
         keep = da.dims[-1]
-    sl = Slicer(da=da, keep=[keep], crop=crop, **kwargs)
-    save_tool = LineSaveTool(data_node=sl.slice_node,
+    sl = Slicer(da, keep=[keep], crop=crop, **kwargs)
+    save_tool = LineSaveTool(data_node=sl.slice_nodes[0],
                              slider_node=sl.slider_node,
-                             fig=sl.fig)
+                             fig=sl.figure)
     from ..widgets import Box
-    return Box([[sl.fig, save_tool.widget], sl.slider])
+    return Box([[sl.figure, save_tool.widget], sl.slider])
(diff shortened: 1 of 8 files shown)


src/plopp/functions/superplot.py
```diff
@@ -66,10 +66,53 @@ def remove_line(self, change: Dict[str, Any], line_id: str):
         self._update_container()
 
 
+class Superplot:
+    """
+    Object that slicing all but one dimension of the input, with a slider per sliced
```
jl-wynen, line 71:
```suggestion
    Object that slices all but one dimension of the input, with a slider per sliced
```

Reads a bit strange. Why 'Object'? Is this a figure or a component of a figure or how is this supposed to be used?

tests/functions/slicer_test.py
```diff
@@ -1,46 +1,75 @@
...
 
 def test_creation_keep_two_dims():
     da = data_array(ndim=3)
-    sl = Slicer(da=da, keep=['xx', 'yy'])
+    sl = Slicer(da, keep=['xx', 'yy'])
     assert sl.slider.value == {'zz': 0}
     assert sl.slider.controls['zz']['slider'].max == da.sizes['zz'] - 1
-    assert sc.identical(sl.slice_node.request_data(), da['zz', 0])
+    assert sc.identical(
+        list(sl.figure.graph_nodes.values())[0].request_data(), da['zz', 0])
```
jl-wynen, line 16: Can we rely on the order of nodes? I.e. is the order part of the contracts of the API?

It seems to me that a relatively high level type like `Slicer` may choose an order internally and change it if that is beneficial or just as a side effect of a refactor.

src/plopp/functions/superplot.py
nvaytet: Mainly, it's just an implementation detail which makes it easier to write unit tests. When running the unit tests, we are not in a jupyter notebook, and figures are thus not widgets. We cannot embed them in the `Box` at the end of the `superplot` function. So it's easier to put most of the code in this class, which can be tested,

The `superplot` function is then just a thin wrapper which is not covered by the unit tests.

It's not meant to be used anywhere outside of the `superplot` function, which is why I struggled to find a good name/description.

tests/functions/slicer_test.py
nvaytet: Hmm there should only be one node here, so order shouldn't matter. But in the other tests I used `sl.slice_nodes[0]` instead, which makes more sense. So I will change to that. Thanks!

src/plopp/functions/superplot.py
nvaytet: I added comments in the docstrings about this.

jl-wynen (approved):

PR #96, opened 2022-11-21
Docs for subplots and figure customization

nvaytet:


docs/customization/subplots.ipynb
```diff
@@ -0,0 +1,148 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "46d49d86-b72b-4083-98ff-6c322c1badca",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "fig, ax = plt.subplots(2, 2, figsize=(8, 6))\n",
+    "\n",
+    "p1 = pp.plot(da['zz', 0], ax=ax[0, 0])\n",
+    "p2 = pp.plot(da['zz', 0]['yy', 0], ax=ax[0, 1])\n",
```
SimonHeybrock, line 54: Not related to this PR, but it is a bit odd that you have `pp.data.data_array` create data arrays that look like meant for testing (with 'xx', 'yy', ... as dim names).

docs/customization/subplots.ipynb
```diff
@@ -0,0 +1,148 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "61f0a00d-fedf-42fb-bf92-303f826dad30",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "size = (4, 3)\n",
+    "\n",
+    "p1 = pp.plot(da['zz', 0], figsize=size)\n",
```
SimonHeybrock, line 96: I was confused here for a couple of seconds, thinking that the `figsize=size` argument took the role of the `ax` argument, telling plopp the "size" of the layout.

Can we avoid specifying the `figsize`? It distracts from what actually needs to be done? Or add an example with 2 plots first, and then give this, explaining how we may want/need to scale down the individual plots?

docs/customization/tweaking-figures.ipynb
```diff
@@ -0,0 +1,196 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a9a8ca49-7df0-4647-96af-8e709c328834",
+   "metadata": {},
+   "source": [
+    "## Adding a figure title\n",
+    "\n",
+    "To add a title to the figure, use the `title` argument:"
```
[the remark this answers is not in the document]
nvaytet: Actually the data array name appears either in the legend in the case of a 1d plot, or next to the colorbar for a 2d plot. Is that something we shoudl change?

docs/customization/subplots.ipynb
nvaytet: yeah, good point, I should probably change the names for the docs, would make more sense.

docs/customization/subplots.ipynb
nvaytet: What about doing
```Py
subplot_size = (4, 3)
p1 = pp.plot(da['zz', 0], figsize=subplot_size)
```
?

docs/customization/tweaking-figures.ipynb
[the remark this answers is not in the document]
SimonHeybrock: Hmm, don't know. Leave it for now?

docs/customization/subplots.ipynb
SimonHeybrock: Or just split the cell? The creation of the plots is really not the main point, so write something like:

```
Given a numbers of plots, p1, ...:
<code for making plots>
we can create a figure with subplots using:
<VBox/HBox code, etc.>
```

nvaytet: I'll do both ;-)

docs/customization/subplots.ipynb
nvaytet: I changed the labels to `x`, `y`, `z` by default, and set them to `xx`, `yy`, `zz` for the unit tests.

SimonHeybrock (approved):

PR #101, opened 2022-11-29
Fix for new scipp version, and update dependencies

nvaytet:


docs/examples/custom-interfaces.ipynb
```diff
@@ -146,7 +146,10 @@
     "in_node = pp.input_node(da)\n",
     "in_node.name = 'Input node'\n",
     "\n",
-    "from scipp.ndimage import gaussian_filter\n",
+    "try:\n",
+    "    from scipp.ndimage import gaussian_filter\n",
+    "except ImportError:\n",
+    "    from scipp.scipy.ndimage import gaussian_filter\n",
```
SimonHeybrock, line 152: I'd probably revert the order, so the "except" branch is not the one taken normally.

SimonHeybrock (approved):

PR #103, opened 2022-11-30
Refactor inspector plot after mpltoolbox improvements

### LINKED ISSUE #102: Bug with inspector points interaction after zooming on figure
To reproduce:
- add points with the `inspector` plot
- zoom in around one of the points
- turn off zoom and try to move the point

This makes a new point instead of grabbing the point.
Grabbing after zooming does work in pure `mpltoolbox`.

nvaytet:
We make a generic `DrawingTool` that can interact with all the tools from `mpltoolbox` (Points, Rectangles, Lines).
The new `on_change` callback was added to `mpltoolbox` in https://github.com/scipp/mpltoolbox/pull/16, and we make use of this in our generic tool.

We then use this for the inspector plot, instead of the custom `EventHandler` class.

In addition, fix #102


src/plopp/widgets/drawing.py
```diff
@@ -0,0 +1,142 @@
...
+        be a figure, or another graph node.
+    value:
+        Activate the tool upon creation if ``True``.
+    **kwargs:
+        Additional arguments are forwarded to the ``ToggleTool`` constructor.
+    """
+
+    def __init__(self, **kwargs):
+        from mpltoolbox import Points
+        super().__init__(tool=partial(Points, mec='w'), icon='crosshairs', **kwargs)
```
SimonHeybrock, line 130: Is inheritance the right solution here? You do not seem to override a significant number of methods. Can we just inject the `tool` and `convert_raw_data` function when creating a `DrawingTool`, instead of inheriting?

nvaytet: i'll try that and see how it works

nvaytet: So would you:
1. make `DrawingTool` inherit from `ToggleTool` and inject the `Points` and `convert_raw_data`
2. convert `DrawingTool` to an `DrawEventHandler`, make `PointsTool` inherit from `ToggleTool` and give it a member `self.event_handler = DrawEventHandler()`
?

SimonHeybrock: Not sure, I didn't suggest to change `DrawingTool`. But if you look at the code now, `PointsTool` is almost just a "constructor" for `DrawingTool`. You could change it to a function that creates a specific `DrawingTool`.

The other things you mention we can consider later, maybe?

SimonHeybrock (approved):

PR #105, opened 2022-12-02
Fix syntax highlighting in docs

nvaytet:
- Remove `tools.PointsTool` in api reference
- downgrade nbconvert to fix syntax highlighting in docs


nvaytet:
> Why are there changes in `base.txt`, etc.? Shouldn't only `docs.txt` affected? We is there no change in `docs.in`? Did you not use `pip-compile-multi`?

It's because I has hacking around.
Now I've done things properly with `pip-compile-multi`

PR #108, opened 2022-12-06
Add possibility to supply `pixel_size` with a unit in `scatter3d` plots

### LINKED ISSUE #93: Default pixel size in pythreejs plotting assumes `unit='m'`?
Using `plot(projection='3d', positions=var_with_mm_unit)` leads to very large pixel sizes. Does it assume the unit is `m` when determining a default size?

nvaytet:
In the same manner as `vmin` and `vmax` for colorbar or axis limits, `pixel_size` can now either be a number or a scalar with a unit.

Fix #93


src/plopp/graphics/point_cloud.py
```diff
@@ -26,15 +26,14 @@ class PointCloud:
...
-            data: sc.DataArray,
-            pixel_size: float = 1,  # TODO: pixel_size should have units
-            opacity: float = 1):
+    def __init__(self,
+                 *,
+                 x: str,
+                 y: str,
+                 z: str,
+                 data: sc.DataArray,
+                 pixel_size: Union[sc.Variable, float] = 1,
```
SimonHeybrock, line 35:
```suggestion
                 pixel_size: Union[sc.Variable, float, int] = 1,
```

src/plopp/graphics/point_cloud.py
```diff
@@ -44,7 +43,17 @@ def __init__(
...
         self._pixel_size = pixel_size
+        if hasattr(self._pixel_size, 'unit'):
+            if len(set([self._data.meta[dim].unit for dim in [x, y, z]])) > 1:
+                raise ValueError(
+                    f'The supplied pixel_size has unit {self._pixel_size.unit}, but '
+                    'the spatial coordinates do not all have the same units. In this '
+                    'case the pixel_size should just be a float with no unit.')
+            else:
+                self._pixel_size = self._pixel_size.to(dtype=float).to(
+                    unit=self._data.meta[x].unit).value
```
SimonHeybrock, line 56: Why `to().to()`?

nvaytet: using a single call now :+1:

PR #109, opened 2022-12-07
Add option to use non-dimension coordinates for axes

nvaytet:
This adds the `coords` argument to `plot` which allows to specify which non-dimension coord to use.
The dim of the coord specified is used to find which coord should be replaced in the input.

Supplied coords can either be just a string (name of existing coord in input data array):
```Py
pp.plot(da, coords=['x2'])
```
or a Variable
```Py
pp.plot(da, coords=[sc.arange('x', 99.)])
```


docs/user-guide/image-plot.ipynb
```diff
@@ -186,6 +186,30 @@
...
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "234f96fc-b05c-42e9-8443-67c8c8865560",
+   "metadata": {},
+   "source": [
+    "## Using a non-dimension coordinate\n",
+    "\n",
+    "By default, Plopp will use the dimension coordinates to set the horizontal position of the points ad the axis labels.\n",
```
SimonHeybrock, line 197:
```suggestion
    "By default, Plopp will use the dimension coordinates to set the horizontal position of the points and the axis labels.\n",
```

docs/user-guide/image-plot.ipynb
```diff
@@ -186,6 +186,30 @@
...
+    "To use a non-dimension coordinate instead, use the `coords` argument."
```
SimonHeybrock, line 198: I am not sure every reader will be familiar with (non)-dimension coords. Have you considered writing it differently, e.g.,

"... for a dimension of name "abc", plopp will use the corresponding coordinate of name "abc". To use a different coordinate, use the `coords` argument.". Should also change the subsection title accordinly.

What if there is no "abc" coord? It is just using an `arange`, right? Maybe worth mentioning as well.

docs/user-guide/line-plot.ipynb
```diff
@@ -248,6 +248,92 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "32317d3d-2f05-417b-9004-8b0b047ab94a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "da = pp.data.data_array(binedges=True)\n",
```
SimonHeybrock, line 269: I don't think this is very didactic. Can you just have a named function in `plopp.data`? Otherwise you unintentionally teach people the syntax (here: the `binedges` keyword arg) of a function that is irrelevant to them. How about `pp.data.histogram_1d()`? Maybe even something with "sample" or "example" in the name?

docs/user-guide/line-plot.ipynb
```diff
@@ -248,6 +248,92 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a87c40bc-5ab6-4310-a476-f8e5f5c0c16d",
+   "metadata": {},
+   "source": [
+    "## Using a non-dimension coordinate\n",
+    "\n",
+    "By default, Plopp will use the dimension coordinates to set the horizontal position of the points ad the axis labels.\n",
+    "To use a non-dimension coordinate instead, use the `coords` argument."
```
SimonHeybrock, line 323: See above.

docs/user-guide/line-plot.ipynb
```diff
@@ -248,6 +248,92 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "bfcd768c-cc4f-4973-ba55-30c4b7da2d73",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "da = pp.data.data_array()\n",
+    "da.coords['x2'] = da.coords['x']**2\n",
+    "pp.plot(da, coords=['x2'])"
```
SimonHeybrock, line 335: Maybe be more didactic to just make the data array? Should not be too much code, or is it?

src/plopp/functions/common.py
```diff
@@ -117,4 +126,24 @@ def preprocess(obj: Union[ndarray, Variable, DataArray],
         out = out[dim, start:start + width + 2]
     if not ignore_size:
         _check_size(out)
+    if coords is not None:
+        renamed_dims = {}
+        if isinstance(coords, str):
+            coords = [coords]
```
SimonHeybrock, line 132: Is this to support `coords='x2'`? If so, why is `coords=x2` not supported as well?

src/plopp/functions/common.py
```diff
@@ -117,4 +126,24 @@ def preprocess(obj: Union[ndarray, Variable, DataArray],
...
+        for dim_or_var in coords:
+            if isinstance(dim_or_var, str):
+                underlying = out.meta[dim_or_var].dims[-1]
+                renamed_dims[underlying] = dim_or_var
+            else:
+                out.coords[dim_or_var.dims[-1]] = dim_or_var
+        out = out.rename_dims(**renamed_dims)
```
SimonHeybrock, line 139: Why are you renaming dims when strings are given?

src/plopp/functions/common.py
```diff
@@ -117,4 +126,24 @@ def preprocess(obj: Union[ndarray, Variable, DataArray],
...
+    for name, coord in out.meta.items():
+        if coord.ndim < 2:
```
SimonHeybrock, line 141: Isn't sorting also an issue for multi-dim coords?

src/plopp/functions/plot.py
```diff
@@ -36,13 +38,21 @@ def plot(obj: Union[VariableLike, ndarray, Dict[str, Union[VariableLike, ndarray
         Aspect ratio for the axes.
     cbar:
         Show colorbar in 2d plots if ``True``.
+    coords:
+        If supplied, use these coords instead of the input's dimension coordinates.
+        The list can contain both strings and ``Variable``s.
```
SimonHeybrock, line 43: Refer to scipp variables? May be unclear otherwise?

src/plopp/functions/plot.py
```diff
@@ -36,13 +38,21 @@ def plot(obj: Union[VariableLike, ndarray, Dict[str, Union[VariableLike, ndarray
...
+        In the case of a string, the coordinate with the corresponding name in the
+        input data array will be used. In the case of a ``Variable``, it will replace
+        the corresponding dimension coordinate directly.
```
SimonHeybrock, line 46: I think this sentence is mostly redundant, the "If supplied, use these coords instead of the input's dimension coordinates." above implies this. Talking too much about dimension-coords is just confusing, I think.

tests/functions/plot_test.py
```diff
@@ -222,3 +222,33 @@ def test_raises_ValueError_when_given_binned_data():
...
+    assert p.canvas._ymax == 3.3 * da.coords['yy'].max().value
+
+
+def test_use_non_dimension_coords_variables():
+    da = data_array(ndim=2, binedges=True)
+    xx2 = 7.5 * da.coords['xx']
+    yy2 = 3.3 * da.coords['yy']
+    p = pp.plot(da, coords=[xx2, yy2])
+    assert p.dims['x'] == 'xx'
+    assert p.dims['y'] == 'yy'
```
SimonHeybrock, line 244: This is really weird. Why should passing strings lead to dim renaming, but passing coords not?

src/plopp/functions/common.py
nvaytet: It was both to support `coords='x2'` but also to avoid weird errors if people do write `coords='x2'` and the iterator starts iterating over each letter in the dim string, saying `dimension 2 is not found` or something like that.
But maybe I should just raise an error if it is not a list?
Or also support `coords=x2`. Not sure what's best.

tests/functions/plot_test.py
SimonHeybrock: Sure, but why are you renaming dims in the other case?

nvaytet: To get the name of the new coord along the x and y axes?

SimonHeybrock: I find this inconsistent, I would have expected
```python
pp.plot(da, coords=['xx2'])
pp.plot(da, coords=[da.coords['xx2']])
```
to give the same result. If not, it may indicate that the API design is not correct?

docs/user-guide/line-plot.ipynb
nvaytet: I'm adding `data1d()`, `data2d()`, `histogram1d()`, `histogram2d()`, `dataset1d()`, `dataset2d()`

docs/user-guide/image-plot.ipynb
nvaytet: done

nvaytet:
@SimonHeybrock did you have any better ideas on how to move forward with this?
```Py
pp.plot(da, coords=['xx2'])
pp.plot(da, coords=[da.coords['xx2']])
```
wouldn't be improved if we used the old syntax:
```Py
pp.plot(da, labels={'xx': 'xx2'})
pp.plot(da, labels={'xx': da.coords['xx2']})
```
because in the second case, you still don't know what to use for the name along the axis (the coord would need to know its own name).

nvaytet:
> So maybe supporting variables here is the wrong approach? The user should add the coord and then plot?

For now I will only support the `coords=['xx2']` syntax. If supporting variables directly is required in the future, we can re-think the whole API, maybe also how those arguments are passed to `scatter3d`.

SimonHeybrock (approved):

PR #110, opened 2022-12-08
Sync toolbar button values with norm and logxy state set by kwargs

nvaytet:
Before:
```Py
pp.plot(da, norm='log')
```
Screenshot at 2022-12-08 13-08-51

After:
Screenshot at 2022-12-08 13-09-01


src/plopp/graphics/interactive.py
```diff
@@ -20,11 +20,18 @@ def __init__(self, *args, **kwargs):
...
+                'panzoom':
+                tools.PanZoomTool(canvas=self._fig.canvas),
+                'logx':
+                tools.LogxTool(self._fig.canvas.logx,
+                               value=self._fig.canvas.xscale == 'log'),
+                'logy':
+                tools.LogyTool(self._fig.canvas.logy,
+                               value=self._fig.canvas.yscale == 'log'),
+                'save':
+                tools.SaveTool(self._fig.canvas.save)
```
SimonHeybrock, line 34: Are we tying plopp more and more to matplotlib with these things? This is violating the "law" of Demeter and  has probably exactly the consequence the "law" was meant to prevent?

nvaytet: If you are refering to `.canvas` and `.xscale`, those are not matplotlib things. The `.canvas` is our own canvas class, and the `.xscale` is a property on that class. It then internally calls Matplotlib's `ax.get_xscale()` method.
So I would say we are ok from that side?

I could abstract away further by making the `logx` and `autoscale` available to the figure (hiding implementation details from the toolbar), but I don't think that would improve things in terms of depending on mpl?

SimonHeybrock: I am fine with this then, from the MPL point of view. Demeter is still something to keep in mind though in the future.

nvaytet:
@SimonHeybrock see latest commit for an update using the widget backend in tests an enabling a inspector test that was being skipped

SimonHeybrock (approved):

PR #113, opened 2022-12-09
Fix for tox 4.0

SimonHeybrock:


.github/workflows/docs.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - run: python -m pip install --upgrade pip
       - run: python -m pip install -r requirements/ci.txt
       - run: |
-          tox --skip-pkg-install -e docs plopp==${VERSION}
+          tox --skip-pkg-install -e docs -- plopp==${VERSION}
```
nvaytet, line 44: I am assuming that this also works for older versions of `tox` because the CI build passed?

SimonHeybrock: Yes.

nvaytet (approved):

PR #124, opened 2022-12-14
Check dims and units of updated values sent to figures, and perform on-the-fly conversions where possible

### LINKED ISSUE #122: Wrong 1-D plots from mapping to items with distinct dimensions
Example 1:
```python
import scipp as sc

x = sc.arange('x', 0, 5, 0.5)
y = sc.arange('y',0, 12,0.7)
dg = sc.Dataset({'a':x,'b':y})
dg.coords['x'] = x
dg.coords['y'] = y
dg.plot()
```
Creates a plot with "y" labeling the axis.

Example2:
```python
import scipp as sc

x = sc.arange('x', 0, 5, 0.5)
y = sc.arange('y',0, 12,0.7)
dg = sc.Dataset({'a':x,'b':y})
dg.plot()
```
Similar problem for items without coords.

nvaytet:
Wth these changes:
- adding new data to an existing figure with dimensions that differ from, or units that are incompatible with, what is already on the figure will raise an exception
- adding new data to an existing figure with units that can be converted to the units of the data already on the figure will perform the conversion on-the-fly

Examples:
```Py
x = sc.array(dims=['x'], values=[1, 2, 3, 4, 5], unit='m')
y = sc.array(dims=['y'], values=[1, 2, 3, 4, 5], unit='m')
pp.plot({'x': x, 'y': y})  # Raises a DimensionError
```

```Py
a = sc.array(dims=['x'], values=[1, 2, 3, 4, 5], unit='m')
b = sc.array(dims=['x'], values=[50., 100., 150., 200., 250.], unit='s')
pp.plot({'a': a, 'b': b})  # Raises a UnitError
```

```Py
a = sc.array(dims=['x'], values=[1, 2, 3, 4, 5], unit='m')
b = sc.array(dims=['x'], values=[50., 100., 150., 200., 250.], unit='cm')
pp.plot({'a': a, 'b': b})  # scales the values of `b` correctly
```
Screenshot at 2022-12-15 13-50-56

Note that in the second example, the order matters: the first entry in the dict decides the unit for the axes.

Fixes #122

### DIFF
--- a/tests/core/utils_test.py
+++ b/tests/core/utils_test.py
@@ -58,29 +58,3 @@ def test_coord_as_bin_edges_non_dimension_coord():
                       })
     result = coord_as_bin_edges(da, key='y', dim='x')
     assert sc.identical(result, sc.linspace('x', 16.5, 21.5, num=6, unit='m'))
-
-
-def test_make_compatible_raises_different_dimension():
-    x = sc.arange('x', 6., unit='m')
-    with pytest.raises(sc.DimensionError):
-        make_compatible(x, dim='y', unit='m')
-
-
-def test_make_compatible_raises_incompatible_unit():
-    x = sc.arange('x', 6., unit='m')
-    with pytest.raises(sc.UnitError):
-        make_compatible(x, unit='s')
-
-
-def test_make_compatible_comverts_compatible_unit():
-    x = sc.array(dims=['x'], values=[1., 4., 9.], unit='s')
-    expected = sc.array(dims=['x'], values=[1000., 4000., 9000.], unit='ms')
-    result = make_compatible(x, unit='ms')
-    assert sc.identical(result, expected)
-
-
-def test_make_compatible_comverts_compatible_unit_integers():
-    x = sc.array(dims=['x'], values=[10, 20, 30, 40, 50], unit='cm')
-    expected = sc.array(dims=['x'], values=[0.1, 0.2, 0.3, 0.4, 0.5], unit='m')
-    result = make_compatible(x, unit='m')
-    assert sc.identical(result, expected)
--- a/tests/core/utils_test.py
+++ b/tests/core/utils_test.py
@@ -1,9 +1,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
 
-from plopp.core.utils import coord_as_bin_edges, make_compatible
+from plopp.core.utils import coord_as_bin_edges
 import scipp as sc
-import pytest
 
 
 def test_coord_as_bin_edges_midpoints_input():
--- a/tests/core/utils_test.py
+++ b/tests/core/utils_test.py
@@ -1,8 +1,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2022 Scipp contributors (https://github.com/scipp)
 
-from plopp.core.utils import coord_as_bin_edges
+from plopp.core.utils import coord_as_bin_edges, make_compatible
 import scipp as sc
+import pytest
 
 
 def test_coord_as_bin_edges_midpoints_input():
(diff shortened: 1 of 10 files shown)


tests/core/utils_test.py
```diff
@@ -57,3 +58,29 @@ def test_coord_as_bin_edges_non_dimension_coord():
...
+    expected = sc.array(dims=['x'], values=[1000., 4000., 9000.], unit='ms')
+    result = make_compatible(x, unit='ms')
+    assert sc.identical(result, expected)
+
+
+def test_make_compatible_comverts_compatible_unit_integers():
+    x = sc.array(dims=['x'], values=[10, 20, 30, 40, 50], unit='cm')
+    expected = sc.array(dims=['x'], values=[0.1, 0.2, 0.3, 0.4, 0.5], unit='m')
+    result = make_compatible(x, unit='m')
+    assert sc.identical(result, expected)
```
SimonHeybrock, line 86: I would now consider this as "bad" test (it may be fine if you are very confident that you will keep this code, otherwise it may simply get in the way when refactoring). Who cares if or what `make_compatible` raises? What we need is to check that trying to plot such data does not produce nonsense (i.e., raises or converts). You have some tests involving figures and nodes, but none that checks if `plot` with an incompatible dataset actually raises.

nvaytet: I'm not sure I follow, are you saying we should just forget about testing any of the internals, anything in core, and just test the top level functions?

SimonHeybrock: Depends:
- If the tests are useful during development, it is worth writing them.
- If it is "internals" but still widely used (maybe not user-facing API, but still somewhat public), then having/keeping those tests is useful.
- In both cases above, it will get in the way during, e.g., refactoring, i.e., a careful balance is needed.
- Unless we have tests for all user facing behavior (top level functions), it is going to break. What good are your internal tests, if in 2 years someone comes and refactors `plot` such that it no longer using those components?

TL;DR I think we have to first and foremost focus on testing "top level" behaviors.

nvaytet: Firstly, I agree that there should be tests on the high level functions like `plot`.

Secondly, my thinking was: put generic functions in a common place, as some other functions may make use of this in the future (and indeed some of the functions in that file have become useful in other places later on).

Now I could only put these functions in the `utils` file when I need them in another place, and keep them next to where they are used until they are needed in more than one place. However, I find that I often forget about such functions I wrote in the past, if they are scattered all over, and I've ended up re-implementing something very similar, only to discover later I had an equivalent function in a different file. So when I think about making a new generic function, I first look in the `utils` file to check if I can't re-use something from there. If the function is used in more than one place, then I think it should have unit tests.

In addition, having tests on the lower-level functions potentially ensures that any new top-level function we add later, similar to plot, is also covered by the tests.

However, writing this message made me think more about things. With my way of putting things in a common file  right from the start, if as you suggest one day `plot` or `figure` are refactored and no longer use that function, I never know whether it is used in another place or not. Therefore it's not immediately obvious (apart from grepping) if the function (and the unit tests) can be removed. I can easily end up with a bunch of functions in there that are used by no-one.

Basically, I am not sure what the right balance is.

SimonHeybrock: Coincidentally, a very similar current question/discussion: https://softwareengineering.stackexchange.com/questions/442852/how-should-factored-out-code-be-tested-as-part-of-the-tdd-refactoring-step

SimonHeybrock: I think it is fine to develop `util` into something more than pure 'implementation details'. In fact you have mostly done that, i.e., you have docstrings, etc. (naming may need more work, as `make_compatible` is meaningless and the concern you voiced above regarding developers not knowing that this function exists is very likely to become true). For such utilities having tests is indeed fine. But, as we seem to agree, that should almost never replace tests of actual behavior of the user-facing (plot) functions.

nvaytet: So I don't think I should remove the tests from `Figure1d` and `Figure2d`, because they are part of the public API (needed when using graph nodes).
Would you then duplicate the tests between `FigureXd` and `plot` (and also `scatter3d`)?

SimonHeybrock: Hmm, is `plot` returning a `FigureXd`? Then maybe that is what could/should be tested (and its properties, i.e., ensure it is setup correctly)?

SimonHeybrock: I suppose `plot` sets some parameters in the `Figure` constructor? You may want to test that none are lost in the transition, etc.

SimonHeybrock (approved):

PR #132, opened 2023-01-04
Fix docs build

nvaytet:
Use local tags instead of github API to get list of releases


docs/version.py
```diff
@@ -1,32 +1,32 @@
...
+    return sorted(versions, reverse=True)
 
 
 class VersionInfo:
 
-    def __init__(self, repo: str, organization: str = 'scipp'):
-        self._releases = _get_releases(repo=repo, organization=organization)
+    def __init__(self):
+        self._releases = _get_releases()
+        print('self._releases', self._releases)
```
SimonHeybrock, line 29: Did you leave the prints in here on purpose?

SimonHeybrock (approved):

docs/version.py
nvaytet: no, just wanted to make sure they contained the right thing before removing them

PR #135, opened 2023-01-05
Refactor the figure to_widget and use a single InteractiveFig

nvaytet:
This reduces duplication that was in `InteractiveFig1d` and `InteractiveFig1d` (matplotlib).


src/plopp/backends/matplotlib/__init__.py
```diff
@@ -21,9 +21,13 @@ def image(self, *args, **kwargs):
...
+            from .interactive import InteractiveFig
+            return InteractiveFig(*args, **kwargs)
+        else:
+            from .static import StaticFig
+            return StaticFig(*args, **kwargs)
 
     def figure2d(self, *args, **kwargs):
-        from .figure import figure2d as fig2d_mpl
-        return fig2d_mpl(*args, **kwargs)
+        return self.figure1d(*args, **kwargs)
```
SimonHeybrock, line 33: This confused me at first. Would be better to have a common internal function that is called by both `figure1d` and `figure2d`?

SimonHeybrock (commented):
Looks good, just one comment:

SimonHeybrock (approved):

PR #136, opened 2023-01-05
improve styling of pan zoom button by creating a custom ToggleButtons

nvaytet:
Before:
Screenshot at 2023-01-05 15-33-35

After:
Screenshot at 2023-01-05 15-33-14


nvaytet:
Merging as I only moved things around and improved styling.

PR #137, opened 2023-01-06
Skip plotly tests if plotly is not installed (for conda build)

nvaytet:
Use `pytest.importorskip` to skip the plotly tests if Plotly is not installed.
This is mainly used in the conda build step, because the tests where Plotly saves figures to disk require the `kaleido` package which is only available on `pip`, thus making a conda build impossible. So we skip all Plotly tests in the conda build.

Also use `boa` (`conda mambabuild`) to speed up conda build step.


tests/backends/plotly/plotly_canvas_test.py
```diff
@@ -7,6 +7,8 @@
 import tempfile
 import os
 
+pytest.importorskip("plotly")
```
SimonHeybrock, line 10: Do you have (non-release) tests that try to run tests without plotly? Otherwise I can see it coming: We will reintroduce imports in some of the plotly modules and a release build will fail because of this. I think we either need to add such a test or move the `importorskip` to the top.

nvaytet: moving the `importorskip` to the top does not help, because it just doesn't run the tests, but it still tries to import the plotly Canvas just after the `importorskip` :-(

nvaytet: But I agree with your concern, that this only runs during a release build.

SimonHeybrock: Can you make another tox env that does not include plotly?

nvaytet: See update

tox.ini
```diff
@@ -12,6 +12,12 @@ basepython = python3.8
 deps = -r requirements/mini.txt
 commands = pytest tests/minimal_plot.py
 
+[testenv:noplotly]
+description = Test that plotly tests are skipped if plotly is not installed
+basepython = python3.8
+deps = -r requirements/noplotly.txt
+commands = pytest tests/backends/plotly
```
SimonHeybrock, line 19: So this will not catch any other tests that depend on plotly outside that dir/file?

nvaytet: Correct. I don't really want to run all the tests twice?

SimonHeybrock: Not if they are slow...?

nvaytet: I think it's overkill. If a conda build fails again in the future, we can revisit.
Running the tests will get slower and slower as we add more. In principle, all the plotly stuff should be isolated to this folder.
I understand that a higher-level function in the future might depend on plotly, but we maybe have to rethink the whole repo organisation at that point anyway.

SimonHeybrock (approved):

PR #142, opened 2023-01-09
Add rectangle selection example to gallery

nvaytet:
Add an example in the gallery where we create a rectangle selection tool and use it to sum along the vertical direction inside the selected region.
Screenshot at 2023-01-10 11-28-25


src/plopp/widgets/drawing.py
```diff
@@ -76,10 +76,10 @@ def make_node(self, artist):
         output_node.name = f'Output node {len(self._output_nodes)}'
         self._output_nodes[nodeid] = output_node
         if self._destination_is_fig:
-            output_node.add_view(self._destination)
+            output_node.add_view(self._destination._fig)
             self._destination.update(new_values=output_node(), key=output_node.id)
-            self._destination.artists[output_node.id].color = getattr(
-                artist, 'color', artist.edgecolor)
+            self._destination.artists[output_node.id].color = artist.color if hasattr(
+                artist, 'color') else artist.edgecolor
```
SimonHeybrock, line 82: Is this something backend specific?

SimonHeybrock (commented):
Looks ok

src/plopp/widgets/drawing.py
nvaytet: No, it depends which shapes you are drawing. A line will have a `color`, while a rectangle will have a `facecolor` and an `edgecolor`.
All this can only be achieved with the matplotlib backend, because `mpltoolbox` is Matplotlib specific. You cannot use `plotly` here.

SimonHeybrock (approved):

PR #143, opened 2023-01-16
Use pre-commit in ci

jl-wynen:


src/plopp/widgets/cut3d.py
```diff
@@ -197,14 +199,18 @@ def _add_cut(self):
...
+
         for n in self._nodes:
             da = n.request_data()
             delta = sc.scalar((self.slider.max - self.slider.min) * self.thickness,
                               unit=self._unit)
             pos = sc.scalar(self.slider.value, unit=self._unit)
             selection = sc.abs(da.meta[self._dim] - pos) < delta
             if selection.sum().value > 0:
-                self.select_nodes[n.id] = node(lambda da: da[selection])(da=n)
+                self.select_nodes[n.id] = node(partial(select, s=selection))(da=n)
```
jl-wynen, line 213: This one is nasty. The old code did not capture `selection` per loop iteration but only the first one. (Found by flake8.) Is it possible to test that this code works with multiple nodes?

nvaytet: False positive? The selection works fine with multiple nodes.

jl-wynen: Ok. But I would still say we leave it as it is now. It is safer than suppressing the check for this line.

nvaytet (approved):

PR #151, opened 2023-02-02
Fix making plot with bin-edge datetime coord and errorbars

### LINKED ISSUE #146: Cannot make 1d plot for data with datetime coord and variances
Coordinate is datetime bin edges, and data has variances:
```Py
time = sc.array(dims=['time'], values=np.arange(np.datetime64('2017-03-16T20:58:17'),
                                                np.datetime64('2017-03-16T21:15:17'), 10))
v = np.random.rand(time.sizes['time']-1)
da = sc.DataArray(data=sc.array(dims=['time'], values=10 * v, variances=v),
                    coords={'time':time})
pp.plot(da)
```
raises
```
UFuncTypeError: ufunc 'add' cannot use operands with types dtype('<M8[s]') and dtype('<M8[s]')
```

This is because [here](https://github.com/scipp/plopp/blob/main/src/plopp/backends/matplotlib/line.py#L178)  we compute the centre of the bins to place the error bars in the middle of the bins. But the operation fails when dtype is datetime.

The code above works if the coord is not bin-edges.

**Note:** this needs to be fixed in both matplotlib and plotly backends.

nvaytet:
Fix #146

### DIFF
--- a/src/plopp/backends/matplotlib/line.py
+++ b/src/plopp/backends/matplotlib/line.py
@@ -112,8 +112,8 @@ class Line:
         }
 
         if data["hist"]:
-            self._line = self._ax.step(data["values"]["x"].values,
-                                       data["values"]["y"].values,
+            self._line = self._ax.step(data['values']['x'],
+                                       data['values']['y'],
                                        label=self.label,
                                        zorder=10,
                                        **{
@@ -121,8 +121,7 @@ class Line:
                                            **kwargs
                                        })[0]
 
-            self._mask = self._ax.step(data['mask']['x'].values,
-                                       data['mask']['y'].values)[0]
+            self._mask = self._ax.step(data['mask']['x'], data['mask']['y'])[0]
             self._mask.update_from(self._line)
             self._mask.set_color(mask_color)
             self._mask.set_label(None)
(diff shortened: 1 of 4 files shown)


src/plopp/backends/matplotlib/line.py
```diff
@@ -175,7 +175,16 @@ def _preprocess_hist(self, data: dict) -> dict:
...
+            if hist:
+                if x.dtype.kind == 'M':
+                    xint = x.astype(int)
+                    xmid = (0.5 * (xint[1:] + xint[:-1])).astype(int)
+                    xvars = np.array(xmid, dtype=x.dtype)
+                else:
+                    xvars = 0.5 * (xint[1:] + xint[:-1])
+            else:
+                xvars = x
+            data["variances"]["x"] = xvars
```
jl-wynen, line 187:
```suggestion
            data["variances"]["x"] = sc.midpoints(x) if hist else x
```
?

jl-wynen (changes requested):

src/plopp/backends/matplotlib/line.py
nvaytet: thanks, didn't remember our `midpoints` worked for datetime. Also, because these are numpy arrays, I got stuck in thinking only about numpy arrays, but I will convert to Variable and back :+1:

jl-wynen: I wasn't aware that this is dealing in numpy arrays. Does this code matter for performance? That is, how often does this run when users, e.g. zoom into a plot or move a slider?

nvaytet: I'm going to try and change it to convert to numpy later, so I can use Scipp for longer. It's not called on zoom, but it is called on slider interactions.

jl-wynen (approved):
I don't really understand in detail what the new code does. But it doesn't look bad either

PR #159, opened 2023-02-13
Add method to convert canvas to ascii art

nvaytet:
Use
```Py
p = pp.plot(da)
p.canvas.to_ascii()
```
Screenshot at 2023-02-13 16-35-20

### DIFF
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -116,6 +116,37 @@ class Canvas:
         from ipywidgets import Image
         return Image(value=fig_to_bytes(self.fig), format='png')
 
+    def to_ascii(self, width: int = 100) -> str:
+        """
+        Convert the underlying Matplotlib figure to ASCII art.
+
+        Parameters
+        ----------
+        width:
+            The line width to be used for the output image.
+        """
+        from PIL import Image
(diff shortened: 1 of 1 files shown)


nvaytet:
I don't know if it's best to return the string, or to print inside the function.

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -116,6 +116,37 @@ def to_image(self):
...
+        height = int(width * ratio)
+        im = im.resize((width, height))
+        # Convert to grayscale
+        im = im.convert("L")
+        # Convert to ASCII
+        pixels = im.getdata()
+        div = 256 / len(ASCII_CHARS)
+        text = "".join([ASCII_CHARS[int(pixel / div)] for pixel in pixels])
+        pixel_count = len(text)
+        return "\n".join(text[i:(i + width)] for i in range(0, pixel_count, width))
```
nvaytet, line 148: I don'tknow if it's best to return the string or print inside the function.

SimonHeybrock:
I don't feel the result is very useful? Are you sure you want to add and maintain code for this feature?

nvaytet:
No it's not very useful, would be better if it would make a@the axes with border characters and actual numbers next to the ticks.

PR #163, opened 2023-02-20
Fix `show_graph`

nvaytet:
The `show_graph` would fail in the case where a node would have two children, and the second child would be discovered as a parent of a node below the first child. This meant that the second child `id` was already in the `inventory` of the graph builder, and would not be added as a child to the top node.

For illustration, this example
```Py
import plopp as pp

a = pp.input_node(0)
a.name = 'a'

b = pp.node(lambda x: x + 1)(a)
b.name = 'b'

c = pp.node(lambda x: x + 2)(a)
c.name = 'c'

d = pp.input_node(4)
d.name = 'd'

e = pp.node(lambda x, y: x + y)(b, d)
e.name = 'e'

f = pp.node(lambda x, y: x + y)(c, d)
f.name = 'f'

pp.show_graph(a)
```
would produce
Screenshot at 2023-02-20 18-58-08
instead of
Screenshot at 2023-02-20 18-58-36

1. `b` would be found as the first child of `a`
2. `e` is found as child of `b`
3. `d` is found as parent of `d`
4. `f` is found as child of `d`
5. `c` is found as parent of `f`
6. back to children of `a`: `c` is already in the inventory, so it does not get added a second time and the connection between `a` and `c` is never made.

To fix, we make an inventory of all nodes and edges separately.


src/plopp/core/graph.py
```diff
@@ -17,30 +17,51 @@ def _make_graphviz_digraph(*args, **kwargs):
...
+        if key not in nodes:
+            if key not in edges:
+                edges[key] = {start.id}
+            else:
+                edges[key].add(start.id)
+            _walk_graph(start=parent,
+                        nodes=nodes,
+                        edges=edges,
+                        views=views,
+                        hide_views=hide_views)
```
jl-wynen, line 46: Do I understand correctly that this is a depth first traversal to both children and parents? So conceptually this is equivalent to this?
```python
for neighbor in start.neighbors():
    # ordered nodes on edge (a -> b)
    a, b = get_edge(start, neighbor)
    add_edge(a, b)
    walk(neighbor)
```

jl-wynen (approved):

PR #164, opened 2023-02-22
Raise exception when multiple 2d inputs are passed to `plot`

### LINKED ISSUE #162: 2-D plot given multiple inputs puts one image on top of the other, hiding the one below
```python
import scipp as sc
import plopp as pp

var1 = sc.ones(dims=['x', 'y'], shape=[200,300])
var2 = var1 * 100

pp.plot({'a':var1, 'b':var2})
```
plots `var2`. I guess it should raise instead?

nvaytet:
Fix #162


SimonHeybrock:
I don't see how that fixes the issue?

nvaytet:
> I don't see how that fixes the issue?

Well I changed the title of the issue. To me, your original issue wasn't one. Both data were being plotted on the figure.

SimonHeybrock:
> > I don't see how that fixes the issue?
> 
> Well I changed the title of the issue. To me, your original issue wasn't one. Both data were being plotted on the figure.

Where I could see only one?

nvaytet:
> Where I could see only one?

Yes. See https://github.com/scipp/plopp/issues/162#issuecomment-1436450378.
The two images are on top of each other, and since they are not transparent, you see only one.

SimonHeybrock:
> > Where I could see only one?
> 
> Yes. See [#162 (comment)](https://github.com/scipp/plopp/issues/162#issuecomment-1436450378). The two images are on top of each other, and since they are not transparent, you see only one.

Yes, that is not good. It needs to be fixed.

nvaytet:
How would you fix that?

SimonHeybrock (approved):

PR #170, opened 2023-03-02
Fix colormapper normalize limits in the case where data vmin == 1.0

nvaytet:
```Py
import numpy as np
import scipp as sc

a = np.array([[1, 2], [3, 4]])
pp.plot(a)
```
would give
Screenshot at 2023-03-02 13-43-08

Note that the lower bound on the colorbar is 0.9, not 1.0 (which it should be).
This is because we were not correctly catching the case where `vmin == normalizer.vmax`, we were only checking for greater than.
So matplotlib was silently changing it to 0.9 before updating the vmax.

### DIFF
--- a/src/plopp/graphics/colormapper.py
+++ b/src/plopp/graphics/colormapper.py
@@ -43,6 +43,13 @@ def _get_cmap(name: str, nan_color: str = None) -> Colormap:
     return cmap
 
 
+def _get_normalizer(norm: str) -> Normalize:
+    """
+    Get an appropriate normalizer depending on the scaling.
+    """
+    return Normalize(vmin=0, vmax=1) if norm == 'linear' else LogNorm(vmin=1, vmax=2)
+
+
 class ColorMapper:
     """
     A class that handles conversion between data values and RGBA colors.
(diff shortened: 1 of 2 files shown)


src/plopp/graphics/colormapper.py
```diff
@@ -194,6 +200,19 @@ def _set_artists_colors(self, key: str = None):
...
+        Synchronize the underlying normalizer limits to the current state.
+        """
+        # Note that the order matters here, as for a normalizer vmin cannot be set above
+        # the current vmax.
+        if self.vmin >= self.normalizer.vmax:
+            self.normalizer.vmax = self.vmax
+            self.normalizer.vmin = self.vmin
+        else:
+            self.normalizer.vmin = self.vmin
+            self.normalizer.vmax = self.vmax
```
jl-wynen, line 214: The two branches are identical. Why is there an `if`?

Do the setters for `vmin` and `vmax` do something on top of setting values?

nvaytet: The comment just above about how the order matters tries to describe it.
Something happens behind the scenes, but I am not sure what.

jl-wynen (approved):

PR #179, opened 2023-03-14
Add camera argument for 3d scatter plots

### LINKED ISSUE #144: The `camera` argument from the old instrument view does not work with Plopp

nvaytet:
Fixes #144 
In addition added the `near` and `far` options for customization.

~@celinedurniak wait until #175 is merged before reviewing~


nvaytet:
hmm good point, I forgot to update the docstring!

src/plopp/graphics/figscatter3d.py
```diff
@@ -46,6 +46,19 @@ class FigScatter3d(BaseFig):
...
+          3 numbers
+        - ``'look_at'``: the point the camera is looking at, as a ``sc.vector`` or a
+          list of 3 numbers
+        - ``'near'``: the distance to the near clipping plane (how close to the camera
+          objects can be before they disappear), as a ``sc.scalar`` or a single number
+        - ``'far'``: the distance to the far clipping plane (how far from the camera
+          objects can be before they disappear), as a ``sc.scalar`` or a single number
+        If values are provided as raw numbers instead of Scipp variables, their unit
+        will be assumed to be the same as the unit of the ``x``, ``y``, and ``z``
+        coordinates.
```
SimonHeybrock, line 61: Docstring duplicated 3 times. Maybe it is worth defining a simple `dataclass` or function with a single docstring?

celinedurniak (approved):

src/plopp/graphics/figscatter3d.py
nvaytet: Yes, that's one of the annoying things about the current implementation (and it also applies to all the other args).
Technically, the `scatter3d` doesn't need to have the args listed, because it just forwards everything to the `FigScatter3d`. However, @jl-wynen and I decided to have them there so that they would show up nicely in the function docs (and also in IDEs).

`FigScatter3d` needs to have them listed, because it needs to distinguish between the args it sends to the `Canvas`, to the `ColorMapper` and the ones it sends to the `PointCloud`.
Finally, the `Canvas` needs to have it because it needs to have it.

I have not managed to come up with an elegant solution to remove this duplication. Would be great if you have an idea.

src/plopp/core/customtypes.py
SimonHeybrock, line 1: What does `customtypes` refer to? It seems odd to have `Camare` buried here if users are supposed to use it.

src/plopp/backends/pythreejs/canvas.py
```diff
@@ -56,7 +47,9 @@ def __init__(
         self._title_text = title
         self._title = self._make_title()
         width, height = self.figsize
-        self._raw_user_camera = camera if camera is not None else {}
+        self._raw_user_camera = Camera(
+            **(camera if camera is not None else {})
+        ).asdict()
```
SimonHeybrock, line 52: Why unpack to dict and put it back into `Camera`? Why is the `asdict` required?

src/plopp/core/customtypes.py
nvaytet: So i thought you wanted to make a special type just to get the docstring to work, but the function would still be called using a dict, e.g. `pp.scatter3d(da, camera={'position': [0, 1, 2], 'far': 1000})`.
What you are saying is that users should create a camera object and pass it to the plot?
Something like
```Py
cam = pp.Camera(position=[0, 1, 2], far=1000)
pp.scatter3d(da, camera=cam)
```
?
I think it would indeed be annoying for users if they have to import the `Camera` from some deeply embedded place, like `pp.core.customtypes.Camera(...)`. So I would only say to go with this if the `Camera` is available in the top level of the plopp module, because we don't want users to have to import all sorts of things (which is why I prefered to just use a python dict).
Is `pp.Camera` something we want?

src/plopp/backends/pythreejs/canvas.py
nvaytet: This made some filtering below easier, but I agree that it's a little weird. I'll refactor, depending on what we decide above.

src/plopp/core/customtypes.py
SimonHeybrock: Yes, I wanted users to use `Camera`. Either top-level, or maybe nested in a submodule specific for the 3d-scatter stuff?

src/plopp/backends/pythreejs/canvas.py
```diff
@@ -40,6 +47,13 @@ def __init__(self, figsize: Tuple[int, int] = (600, 400), title: str = None):
         self._title_text = title
         self._title = self._make_title()
         width, height = self.figsize
+        if camera is not None:
+            if isinstance(camera, dict):
+                self._user_camera = Camera(**camera)
```
SimonHeybrock, line 52: Is this for legacy compatiblity?

SimonHeybrock (approved):

src/plopp/backends/pythreejs/canvas.py
nvaytet: Yes, but I don't mind removing it.

SimonHeybrock: Remove it then

SimonHeybrock (approved):

PR #189, opened 2023-03-29
Improvements to syntax for using nodes and better graph visualization

nvaytet:
## Syntax changes for `plopp.node` and deprecation of `input_node`

The main change here is that the `plopp.node` decorator no longer accepts kwargs, only a callable.
Instead, kwargs are set in the call to the `Node` constructor itself.

Any input to the `Node` constructor that is not a `Node` gets converted to a `Node` that wraps `lambda: input`.
This means that the special `pp.input_node` is obsolete, and I have added a deprecation warning.

Both of these mean that the syntax is now much closer to `functools.partial`.
For example, to make a node that sums the input data array along the `y` dimension, we currently had to write something like
```Py
sum_node = pp.node(sc.sum, dim='y')(input_node)
```
It was not immediately clear to new users that inside the first set of brackets would go parameters that are normal python objects, and in the second set of bracket would go other nodes in the graph.

The new syntax is
```Py
sum_node = pp.Node(sc.sum, input_node, dim='y')
```
The user no longer has to worry about what goes where.

## Node operators

In addition, I have added simple operators to nodes that allow us to do things like
```Py
signal = pp.Node(input_signal)
noise = pp.Node(background_noise)
clean = signal - noise
clean()
```
Note that `clean` is a `Node`.

## Improved automatic naming of nodes

Finally, I have improved the names that automatically get add inside the graph visualization `show_graph`. Before, if node names were not set, the unique ID of the node was displayed in the graph.
Now, the following code
```Py
import plopp as pp
from scipp.scipy.ndimage import gaussian_filter

data_node = pp.Node(da)
smooth_node = pp.Node(gaussian_filter, data_node, sigma=5)
fig = pp.figure2d(smooth_node)

pp.show_graph()
```
produces
Screenshot at 2023-03-29 16-26-40


src/plopp/core/helpers.py
```diff
@@ -0,0 +1,49 @@
...
+
+from typing import Callable
+
+from .node import Node
+
+
+def node(func: Callable) -> Callable:
+    """
+    Create a :class:`Node` from a callable.
+    The additional arguments will be parent nodes to the returned node.
```
jl-wynen, line 12: What additional arguments?

src/plopp/core/node.py
```diff
@@ -2,10 +2,35 @@
...
+def input_node(obj: Any):
+    """
+    Create a simple node that returns the supplied object when data is requested from
+    it. This node has no parents, and typically lives at the top of a graph to provide
+    the raw input data.
+
+    Parameters
+    ----------
+    obj:
+        The object to return when data is requested from the node.
```
jl-wynen, line 22: Can you add https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated?

src/plopp/core/node.py
```diff
@@ -24,18 +49,35 @@ class Node:
     """
 
     def __init__(self, func: Callable, *parents, **kwparents):
```
jl-wynen, line 51: The type hint doesn't match. `func` can be non-callable. It should just be `Any` now.

src/plopp/core/node.py
```diff
@@ -24,18 +49,35 @@ class Node:
...
+        if func_is_callable:
+            self.func = func
+        else:
+            self.func = lambda: func
+        self._id = uuid.uuid4().hex
         self.children = []
         self.views = []
-        self.parents = list(parents)
-        self.kwparents = dict(kwparents)
+        self.parents = [p if isinstance(p, Node) else Node(p) for p in parents]
```
jl-wynen, line 60: In Dask,
```python
d = Delayed(...)
assert d is delayed(d)
```
note the different capitalisation.
If we do this, too, the code here can be simplified to avoid `isinstance` checks.

src/plopp/core/node.py
```diff
@@ -24,18 +49,35 @@ class Node:
...
     def __init__(self, func: Callable, *parents, **kwparents):
-        if not callable(func):
-            raise ValueError("A node can only be created using a callable func.")
-        self.func = func
-        self._id = str(uuid.uuid1())
+        func_is_callable = callable(func)
+        if func_is_callable:
+            self.func = func
+        else:
+            self.func = lambda: func
```
jl-wynen, line 56: Writing the conversion to dask graphs, I wish I had a way to get the input value out of an input node. Can you store `func` in an attr like `_input_data` if it is not callable? Then properties like `input_data` (getter only) and `is_input_node` would help to inspect the graph.

docs/examples/custom-interfaces.ipynb
jl-wynen, line 1: Does it make sense to also automatically convert to nodes in the `figurend` functions? Then we can get rid of most calls or all to `pp.Node` here.

nvaytet: I am not sure I want to go that far. I think this would almost make the `figure` behave like the `plot` wrapper.
I think it is cleaner to know that figures only accept nodes as inputs.

jl-wynen: ok

src/plopp/core/node.py
nvaytet: Then you would have something like
```Py
self.parents = [p if p is node(p) else Node(p) for p in parents]
```
?
Not sure it's worth it, using standard Python `isinstance` seems ok to me?

src/plopp/core/node.py
nvaytet: Done.

src/plopp/core/node.py
jl-wynen: No, I mean you can use `node` with a `Node` instance and it returns the same object. So you could write `self.parents = [node(p) for p in parents]`.

src/plopp/core/node.py
nvaytet: It shows up fine if i put it before the `Parameters` section:
Screenshot at 2023-03-30 14-27-47

src/plopp/core/node.py
nvaytet: Would it be the lowercase (decorator) node, or the constructor uppercase `Node`?
`self.parents = [Node(p) for p in parents]` ?

jl-wynen: The lowercase decorator. Let's not make a constructor return the input object!

jl-wynen: Now I get it. `node` returns a 'node constructor' not an actual node. Maybe the name isn't very good because
```python
a = node(f)
b = Node(f)

a() # creates a node but does not run any computation
b() # computes a result
```

nvaytet: Yeah, but I wanted users to not have to remember too many new names.
I think it kind of makes sense when it's used as a decorator?
```Py
@node
def add(x, y):
    return x + y

my_adding_node = add(input_node(a), input_node(b))
```
The decorator is turning the add function into a node, kind of.

jl-wynen: > Well something that makes a node anyway.

This is my point.
I'm ok with leaving it as is. But I'm curious if this trips up other people as well.

jl-wynen (approved):

PR #209, opened 2023-05-24
Change the behaviour of autoscaling limits

### LINKED ISSUE #178: Behaviour of autoscaling colorbar
Until now, the limits on the colorbar of a 2d plot are allowed to grow with time, but not shrink.
This means that if a figure receives new data, e.g. a new slice selected from moving a slider, if the range of new values is wider than the current range, the limits on the colorbar will autoscale to the new data.
If the range is smaller, the colorscale will not change.

This was intentional to give a better feel for the relative intensities of the different slices, while at the same time allowing the user to see new features as they appear.

However, this does have the strange effect that if a user is navigating through a 3d data set, sees a feature (some peaks), continues through the data, and then wants to come back to the peaks, the colorscale may now be different and the peaks could have been washed out if a higher range was displayed when the slider when past the peaks.

The two alternatives are:
1. Use the whole range of the 3d data as the limits, and keep them constant. This may mean that the user never sees those small peaks in the first place, but it has the advantage of moving the slider back to a previous location will always show the same image
2. Always autoscale on the displayed slice. This may be what the user expects the most, and would also show the same image when going back to a previous slider position. But the entire custom mechanism was designed to provide a better feel for the data.

A third option could be to keep the current mechanism, but then use the `Home` button to both autoscale the axes and the colorscale. Currently the `Home` button just goes back to the currently set limits, which may be wider than the displayed data, if the slider previously took you to a region where the range of values was wider.

@celinedurniak

nvaytet:
By default, limits on the vertical axis (1d plots) and the colorbar (2d plots) will now fit to the currently displayed data, instead of having 'memory' and only be allowed to grow.
The old behaviour can be recovered by doing `autoscale='grow'`.

Fix #178

### DIFF
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -141,10 +141,6 @@ class Canvas:
         draw:
             Make a draw call to the figure if ``True``.
         """
-        xmin = np.inf
-        xmax = np.NINF
-        ymin = np.inf
-        ymax = np.NINF
         if self.ax.lines:
             self.ax.relim()
             self.ax.autoscale()
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -46,6 +46,10 @@ class Canvas:
         The maximum value for the vertical axis. If a number (without a unit) is
         supplied, it is assumed that the unit is the same as the current vertical axis
         unit.
+    autoscale:
+        The behavior of the axis limits. If ``auto``, the limits automatically
+        adjusts every time the data changes. If ``grow``, the limits are allowed to
+        grow with time but they do not shrink.
     aspect:
         The aspect ratio for the axes.
     cbar:
@@ -61,6 +65,7 @@ class Canvas:
         grid: bool = False,
         vmin: Union[sc.Variable, int, float] = None,
         vmax: Union[sc.Variable, int, float] = None,
+        autoscale: Literal['auto', 'grow'] = 'auto',
         aspect: Literal['auto', 'equal'] = 'auto',
         cbar: bool = False,
         **ignored,
@@ -82,6 +87,7 @@ class Canvas:
         self.xunit = None
         self.yunit = None
         self._own_axes = False
+        self._autoscale = autoscale
 
         if self.ax is None:
             self._own_axes = True
@@ -135,15 +141,24 @@ class Canvas:
         draw:
             Make a draw call to the figure if ``True``.
         """
+        xmin = np.inf
+        xmax = np.NINF
+        ymin = np.inf
+        ymax = np.NINF
         if self.ax.lines:
             self.ax.relim()
             self.ax.autoscale()
             xmin, xmax = self.ax.get_xlim()
             ymin, ymax = self.ax.get_ylim()
-            self._xmin = min(self._xmin, xmin)
-            self._xmax = max(self._xmax, xmax)
-            self._ymin = min(self._ymin, ymin)
-            self._ymax = max(self._ymax, ymax)
+            # self._xmin = min(self._xmin, xmin)
+            # self._xmax = max(self._xmax, xmax)
+            # self._ymin = min(self._ymin, ymin)
+            # self._ymax = max(self._ymax, ymax)
+        else:
+            xmin = np.inf
+            xmax = np.NINF
+            ymin = np.inf
+            ymax = np.NINF
         for c in self.ax.collections:
             if isinstance(c, QuadMesh):
                 coords = c.get_coordinates()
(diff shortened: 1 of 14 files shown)


src/plopp/backends/matplotlib/canvas.py
```diff
@@ -46,6 +46,10 @@ class Canvas:
         The maximum value for the vertical axis. If a number (without a unit) is
         supplied, it is assumed that the unit is the same as the current vertical axis
         unit.
+    autoscale:
+        The behavior of the axis limits. If ``auto``, the limits automatically
+        adjusts every time the data changes. If ``grow``, the limits are allowed to
+        grow with time but they do not shrink.
```
SimonHeybrock, line 52: Would a third option make sense: Computing vmin/vmax and setting a fixed scale. I know we are not doing this automatically due to the cost (even though I am not sure about that argument any more --- it was indeed a problem with event data and dynamic resampling, but is it still an actual problem?), but having it as a "costly" non-default option may be useful?

Edit: I now realize that this may not be possible in the Plopp architecture, as the plot does never see the higher-dimensional data in case of `slicer`. So maybe `slicer` should have that additional `autoscale` option (and set `vmin/vmax` based on that?

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -135,15 +141,20 @@ def autoscale(self):
...
             ymin, ymax = self.ax.get_ylim()
-            self._xmin = min(self._xmin, xmin)
-            self._xmax = max(self._xmax, xmax)
-            self._ymin = min(self._ymin, ymin)
-            self._ymax = max(self._ymax, ymax)
+        else:
+            xmin = np.inf
+            xmax = np.NINF
+            ymin = np.inf
+            ymax = np.NINF
```
SimonHeybrock, line 157: Setting this twice?

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -159,10 +170,20 @@ def autoscale(self):
...
                 )
-                self._xmin = min(self._xmin, left.value)
-                self._xmax = max(self._xmax, right.value)
-                self._ymin = min(self._ymin, bottom.value)
-                self._ymax = max(self._ymax, top.value)
+                xmin = min(xmin, left.value)
+                xmax = max(xmax, right.value)
+                ymin = min(ymin, bottom.value)
+                ymax = max(ymax, top.value)
+        if self._autoscale == 'grow':
```
SimonHeybrock, line 177: Why does `autoscale` (which is for the colorbar) affect the behavior for X and Y limits?

src/plopp/graphics/colormapper.py
```diff
@@ -166,39 +172,37 @@ def rgba(self, data: sc.DataArray) -> np.ndarray:
...
+        Re-compute the global min and max range of values by iterating over all the
+        artists. If autoscale is set to ``'auto'``, the limits adjust to he current
+        range. If it is set to ``'grow'``, limits can grow but not shrink.
         """
-        vmin, vmax = fix_empty_range(find_limits(data, scale=self.norm))
+        limits = [
+            fix_empty_range(find_limits(artist._data, scale=self.norm))
+            for artist in self.artists.values()
+        ]
+        vmin = reduce(lambda a, b: min(a, b), [v[0] for v in limits])
```
SimonHeybrock, line 185: Isn't this the same as
```suggestion
        vmin = reduce(min, [v[0] for v in limits])
```
?

src/plopp/backends/matplotlib/canvas.py
nvaytet: Yes, it's the reason you list in your edit. We could add it to `slicer`, i'd be happy with that.

src/plopp/graphics/colormapper.py
nvaytet: ah man, let me unnecessarily over-complicate things, your suggestion is too clean for me... ;-)

src/plopp/backends/matplotlib/canvas.py
nvaytet: Yes I also noticed that yesterday :+1:

src/plopp/functions/slicer.py
```diff
@@ -68,6 +72,16 @@ def __init__(
                 f"were not found in the input's dimensions {ds.dims}."
             )
 
+        if autoscale == 'fixed':
+            if ('vmin' in kwargs) or ('vmax' in kwargs):
+                raise ValueError(
+                    'Slicer plot: when autoscale is set to "fixed", vmin and/or vmax '
+                    'cannot be specified.'
+                )
```
SimonHeybrock, line 90: Not sure this is necessary or convenient? Say I want to first see the full range, then manually set a maximum, or play with the settings. I think the `autoscale` could still apply to the other (vmin or vmax) that was not user-provided. If both are provided, maybe just warn, instead of raising?

Otherwise you could argue that also the `auto` or `grow` settings should raise if manual limits are given.

src/plopp/functions/slicer.py
```diff
@@ -83,13 +84,16 @@ def __init__(
...
+            if None not in (vmin, vmax):
+                warnings.warn(
+                    'Slicer plot: autoscale is set to "fixed", but vmin and vmax '
+                    'are also specified. They will override the autoscale setting.',
+                    RuntimeWarning,
                 )
-            vmin = reduce(min, [da.data.min() for da in ds.values()])
-            vmax = reduce(max, [da.data.max() for da in ds.values()])
+            if vmin is None:
+                vmin = reduce(min, [da.data.min() for da in ds.values()])
```
SimonHeybrock, line 94: Note that scipp should be able to do this via
```suggestion
                vmin = sc.reduce([da.data.min() for da in ds.values()]).min()
```
but what you have is also fine.

SimonHeybrock (approved):

PR #212, opened 2023-05-30
Add hover values for 2d image plots

### LINKED ISSUE #47: Missing cursor hover values in 2d image plots
This is because we are now using `pcolormesh` instead of `imshow`, which used to give the value by default.

We should try to re-enable this, probably using something like https://stackoverflow.com/a/42581480/13086629

We could either have this enabled/disabled by a 'hover tool' in the toolbar, or have it enabled by default, as long as the number of pixels is small (because the searchsorted should be fast enough), and have as an opt-in for larger images.

nvaytet:
Fix #47 

This is enabled by default.
I have not yet decided if it should be activated via a button in the toolbar.
I've tried on a few images and it seemed relatively fast.


src/plopp/backends/matplotlib/image.py
```diff
@@ -130,6 +132,31 @@ def __init__(
...
+                    self._data.coords[self._dim_1d[1]].unit,
+                    self._data.coords[self._dim_2d[1]].unit,
+                ),
+            }
+        else:
+            self._hover_slicing = {
+                'dir': ('y', 'x'),
+                'dim': (self._data.dims[0], self._data.dims[1]),
+                'unit': (
+                    self._data.coords[self._data.dims[0]].unit,
```
SimonHeybrock, line 153: dim 0 is Y?

src/plopp/backends/matplotlib/image.py
```diff
@@ -175,3 +202,37 @@ def update(self, new_values: sc.DataArray):
...
+        data at that point.
+
+        Parameters
+        ----------
+        x:
+            The x coordinate of the mouse pointer.
+        y:
+            The y coordinate of the mouse pointer.
+        """
+        out = f'x={x:1.4f}, y={y:1.4f}'
```
SimonHeybrock, line 218: Does this format in a meaningful manner for very large or very small ranges? Do we need to make the format depend on the min/max of the axis?

src/plopp/backends/matplotlib/image.py
```diff
@@ -175,3 +202,37 @@ def update(self, new_values: sc.DataArray):
...
+                ),
+            ]
+            val = val[
+                self._hover_slicing['dim'][1],
+                sc.scalar(
+                    xy[self._hover_slicing['dir'][1]],
+                    unit=self._hover_slicing['unit'][1],
+                ),
+            ]
+            out += f', z={val.value:1.4f}'
```
SimonHeybrock, line 235: Same question, do we need different formatting, e.g., scientific?

Further, should we include the units, I think it would improve the UX?

Can we make use of the "compact" variable formatting?

nvaytet: Good point about the unit. Do we also want to add the units for the cursor values in 1d plots?

SimonHeybrock: Yes.

nvaytet:
It now does this:
Screenshot at 2023-05-31 13-22-40

src/plopp/backends/matplotlib/image.py
SimonHeybrock: Yes, probably.

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -237,6 +238,23 @@ def crop(self, **limits):
...
+        Parameters
+        ----------
+        x:
+            The x coordinate of the mouse pointer.
+        y:
+            The y coordinate of the mouse pointer.
+        """
+        return (
+            f'x={scalar_to_string(x * self.xunit)}, '
+            f'y={scalar_to_string(y * self.yunit)}'
```
SimonHeybrock, line 255: Shouldn't it use the dim name instead of x and y?

src/plopp/backends/matplotlib/image.py
```diff
@@ -215,7 +215,10 @@ def format_coord(self, x: float, y: float) -> str:
         y:
             The y coordinate of the mouse pointer.
         """
-        out = f'x={x:1.4f}, y={y:1.4f}'
+        out = (
+            f'x={scalar_to_string(x * self._hover_slicing["unit"][1])}, '
+            f'y={scalar_to_string(y * self._hover_slicing["unit"][0])}'
```
SimonHeybrock, line 220: Dito

src/plopp/backends/matplotlib/image.py
```diff
@@ -232,7 +235,7 @@ def format_coord(self, x: float, y: float) -> str:
                     unit=self._hover_slicing['unit'][1],
                 ),
             ]
-            out += f', z={val.value:1.4f}'
+            out += f', z={scalar_to_string(val)}'
```
SimonHeybrock, line 238: I think `z` is confusing, especially since you could have a dimension labelled `z`. Can we indicate in some other way that this is the value?

Some ideas:
```
f(time=1 s, temp=1 K) = 100 counts
(time=1 s, temp=1K) -> 100 counts
```

nvaytet: I went with `(time=1 s, temp=1 K) : 100 counts`

src/plopp/backends/matplotlib/image.py
```diff
@@ -235,7 +237,7 @@ def format_coord(self, x: float, y: float) -> str:
                     unit=self._hover_slicing['unit'][1],
                 ),
             ]
```
SimonHeybrock, line 239: I don't know if it is noticeable as lag, but this function does quite a few things (such as allocating new variables, dict and list accesses, some f-string bits) that could be avoided. Have you considered moving those bits to a place where it is done only once? Here you would then only, e.g., set the x and y values in the existing variables, format the remaining bits, and to the index lookup.

src/plopp/backends/matplotlib/line.py
```diff
@@ -221,6 +222,24 @@ def update(self, new_values: sc.DataArray):
...
+        ----------
+        x:
+            The x coordinate of the mouse pointer.
+        y:
+            The y coordinate of the mouse pointer.
+        """
+        vx = sc.scalar(x, unit=self._canvas.xunit)
+        vy = sc.scalar(y, unit=self._canvas.yunit)
+        return (
+            f'{self._canvas.xdim}={scalar_to_string(vx)} : ' f'{scalar_to_string(vy)}'
```
SimonHeybrock: So this is not actually displaying the value, just the mouse-y? Then I think this `f(...): ...` syntax is very misleading (unlike in the 2-d case)

nvaytet: So you are saying you don't like that it is not snapping to the closest value?
I guess this is difficult in the case of multiple lines. We don't really want one entry per line I think, because that may get very expensive with 20+ lines (which can easily happen when you use e.g. `sc.collapse`).

But that does raise a good point about the 2d images. Currently, although not common, you can have more than one image on the same axes (we prevent this in the `plot` function cf a discussion we had a while back), but it is still possible via the lower-level `FigImage`. If the `axformatter` is specified by the `Image` artist, when you add a new image it will override the formatter from the first image and only display values when you hover over the second image, not the first one.
So maybe I should have a global formatter on the canvas, and a mechanism where the artist register additional formatters.

For the 1d case, I am thinking that we should just display x and y and not try to find the closest value?

SimonHeybrock: > So you are saying you don't like that it is not snapping to the closest value?

No, I am saying we either have to snap to the closest value (and keep the display as it is), or change the display. As it is now the user may *think* that it displays the closest value, which is dangerous.

nvaytet: I think i'd rather change the display to show the cursor x,y values. But I guess we still want the dim label and the units in there?
`time=40 s, 67 counts` would look weird though.
And we didn't want to use `x=40s, y=67 counts` because the horizontal axis may have `y` as a dimension, making things confusing...

SimonHeybrock: Hmm, maybe the dim label can be omitted here, since there is only one? `40 s, 67 counts`?

SimonHeybrock: Somehow I fear that even with this notation, sooner or later a user will think it is the actual value, not the mouse position. Can we make it even clearer?

SimonHeybrock: Can we display actual values, displaying a dict of values if there are multiple lines?

```
time:40 s -> (sample: 10 counts, background: 100 counts)
```

SimonHeybrock: I suppose that still does not solve the problem of which X value to use, since doing an interpolation seems wrong and too complicated.

nvaytet: > Can we display actual values, displaying a dict of values if there are multiple lines?

We could, but not with 20 lines

src/plopp/backends/matplotlib/image.py
nvaytet: I tried to improve this slightly.

SimonHeybrock:
> This is what I came up with in the end:
>     * When there is just one 2d image: `(time=24 s, temperature=300 K): {sample: 152 counts}`

I find it quite odd to have the dict notation, since in 99% of cases there is only one image?

SimonHeybrock:
So will it look oddly broken if the data array has no name?

nvaytet:
It looks like this: `(time=24 s, temperature=300 K): {152 counts}`.
I can add more logic to remove the braces if there is no name, but it is a trade-off between complexity and how good it looks...

SimonHeybrock (approved):

PR #216, opened 2023-06-06
Raise in slicer when too many keep dims are requested

nvaytet:
Before these changes, if you requested 3 dims in `keep`, you would get this strange error
```
AttributeError: 'Slicer' object has no attribute 'figure'
```

### DIFF
--- a/src/plopp/functions/slicer.py
+++ b/src/plopp/functions/slicer.py
@@ -105,7 +105,8 @@ class Slicer:
         self.slice_nodes = [
             slice_dims(data_node, self.slider_node) for data_node in self.data_nodes
         ]
-        if len(keep) == 1:
+        ndims = len(keep)
+        if ndims == 1:
             self.figure = figure1d(
                 *self.slice_nodes,
                 crop=crop,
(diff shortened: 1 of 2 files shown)


src/plopp/functions/slicer.py
```diff
@@ -105,7 +105,8 @@ def __init__(
         self.slice_nodes = [
             slice_dims(data_node, self.slider_node) for data_node in self.data_nodes
         ]
-        if len(keep) == 1:
+        ndims = len(keep)
+        if ndims == 1:
```
jl-wynen, line 109: Both branches pass the same args to `figure1d` and `figure2d`. You can avoid this, e.g., by using
```python
if ndims == 1:
    make_figure = figure1d
# ...
self.figure = make_figure(...)
```
or something similar.

jl-wynen (approved):
Just a really minor comment

src/plopp/functions/slicer.py
nvaytet: I like minor comments ;-)

jl-wynen (approved):

PR #218, opened 2023-06-13
Fix plotly

### LINKED ISSUE #217: Plotly backend is broken
Since #212 , the `plotly` backend is broken: it raises with `empty` is not defined on the canvas.

nvaytet:
Fixes #217

### DIFF
--- a/src/plopp/backends/manager.py
+++ b/src/plopp/backends/manager.py
@@ -1,10 +1,12 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
+DEFAULTS = {'2d': 'matplotlib', '3d': 'pythreejs'}
+
 
 class BackendManager:
     def __init__(self):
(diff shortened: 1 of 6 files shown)


src/plopp/backends/manager.py
```diff
@@ -24,6 +26,10 @@ def values(self):
     def items(self):
         return self._mapping.items()
 
+    def reset(self):
+        self._mapping = DEFAULTS.copy()
```
jl-wynen, line 30: Why does it store `DEFAULTS` as an instance variable? And why does it reset it here?

nvaytet: I'm not sure I understand the question. Resetting here inside the `reset` method makes sense?
Are you saying I should do it differently, something like
```Py
self._mapping.clear()
self._mapping.update(DEFAULTS)
```
?

jl-wynen: Ah sorry. I got confused about `_mapping` and `_backends` and thought that `_mapping` never changes.

jl-wynen (approved):

PR #220, opened 2023-06-29
Add tiled plots

### LINKED ISSUE #180: Adding operators to plots?
Some projects like Holoviz have added operators to plots (see e.g. [here](https://holoviz.org/tutorial/Composing_Plots.html)).
For example, they use the `*` operators between two plots to overplot the data from both onto the same axes.
They use the `+` operator to make tiled plots.

Do we want to add something similar?
While I am not sure about the usefulness of the `*` operator (and also not sure whether it is a good choice), maybe some aspects are interesting? Making tiled plots with `p1 + p2` is definitely appealing, instead of having to always use the `Box([p1, p2])`.
I don't know how you would make horizontal and vertical tiles. Maybe the `/` operator could be used to make one plot above the other.
So something with 4 panels would be `(p1 + p2) / (p3 + p4)`.

I think having operators would definitely be beneficial for cropping (setting the axes limits).
Right now, to set the axis limits, one needs to use a rather strange syntax to set the limits for the `time` dimension:
`da.plot(crop={'time': {'min': 2.5, 'max': 40.}})`
It would be nice if instead we could use the same slicing syntax as we have on sc.DataArray to crop:
`da.plot()['time', 2.5:40.]`.

Note however that using integers may become confusing, as `da.plot()['time', 2:40]` would be cropping on the values of the `time` coordinate, not slicing using integer indices.

nvaytet:
Examples:
```Py
import plopp as pp

p1 = pp.plot(pp.data.data1d(), grid=True, title="My Title", vmax=10)
p2 = pp.plot(pp.data.data2d(variances=True))

tiled = pp.tiled(nrows=2, ncols=2)
tiled[0, 0] = p1
tiled[0, 1] = p2
tiled[1, 0] = p2
tiled[1, 1] = p1
tiled
```
```Py
tiled = Tiled(nrows=1, ncols=3)
tiled[0, :2] = p1
tiled[0, 2] = p2
tiled
```

There are also operators on figures:
```Py
p1 + p2  # side by side plots
p1 / p2  # vertically stacked plots
(p1 + p2) / (p2 + p1)  # 2x2 figure
```

Fixes #180


nvaytet:
I'm struggling a bit with implementing this one.

The issue I have is the following.
We imagined for tiled plots, we could do
```Py
tiled = pp.tiled(1, 2)
tiled[0] = da1.plot()
tiled[1] = da2.plot()
tiled
```

which would also mean that one could do

```Py
p1 = da1.plot()
p2 = da2.plot()
tiled = pp.tiled(1, 2)
tiled[0] = p1
tiled[1] = p2
```

Now the way I've currently implemented tiled plots to be created is that when you add a plot to a subplot of the `tiled` object, it clears the figure of the current plot `p1`, and re-assigns the axes of the plot's `canvas` to be the new tiled axes, and then re-draws the current data in the plot onto the new axes.
This was done because I thought that if you then later want to update the plot (e.g. action from a slider), and the plot is now part of a tiled plot, then it should still update there.

The issue is now that `p1` has an empty figure, so while the repr from `tiled` will show the plot correctly, `p1`'s repr will just be empty.
I am struggling to find a way to have two MPL figures (one in tiled and one in `p1`) which stay in sync when the data is updated.

This is making me feel that the design is flawed in a way, and that we should maybe use the subplot axes when the plot is created, not add an existing plot to tiled subplots.
What I mean is it would be more like
```Py
tiled = pp.tiled(1, 2)
da1.plot(ax=tiled[0])
da2.plot(ax=tiled[1])
tiled
```
It's not as nice I think, but would solve the issue above.

However, I liked the design above because I could add operators to plots and do `p1 + p2` to make 2 plots side by side (uses `pp.tiled` internally).
But this once again has the issue that after i've done `p1 + p2`, if I then want to look at the repr of `p1` further down in the notebook, that has been silently swallowed by the `+` operation and is now empty, which is very strange.....

nvaytet:
In the end, I am now making copies of plots when they are inserted into a tiled plot.
This means that doing
```Py
tiled = pp.tiled(1, 2)
tiled[0] = p1
tiled[1] = p2
```
and then `p1.update(new_values)` will only update `p1`, not `tiled[0]`.

This is a little non-pythonic if we think of `tiled` as being a bit like a dict, but has the advantage of not having to transfer axes, update reprs, etc... and you can still view `p1`'s repr after the tiled construction.

nvaytet:
Hmm, I still need to make this work with Plotly.

nvaytet:
> Hmm, I still need to make this work with Plotly.

Getting this to work with Plotly is not super easy. So I opened a new issue #223 , and will defer to a later PR.

SimonHeybrock:
> What I mean is it would be more like
> 
> ```python
> tiled = pp.tiled(1, 2)
> da1.plot(ax=tiled[0])
> da2.plot(ax=tiled[1])
> tiled
> ```
> 
> It's not as nice I think, but would solve the issue above.

Isn't this close to how Matplotlib does it? In that case, I'd :+1: this.

nvaytet:
> Isn't this close to how Matplotlib does it? In that case, I'd +1 this.

Not really, with mpl, you have to do
```Py
fig = plt.figure()

gs = GridSpec(2, 2, figure=fig)
ax1 = fig.add_subplot(gs[0, 0])
ax2 = fig.add_subplot(gs[0, 1])
ax3 = fig.add_subplot(gs[1, 0])
ax4 = fig.add_subplot(gs[1, 1])
ax1.plot(x1, y1)
ax2.plot(x2, y2)
ax3.imshow(image_data)
ax4.scatter(x3, y3)
```
You first create the axes, and then use them to plot. You don't pass the axes to the plotting functions.

src/plopp/backends/matplotlib/tiled.py
```diff
@@ -0,0 +1,206 @@
...
+        self.axes = []
+        self.views = np.full((nrows, ncols), None)
+        self._history = []
+
+    def __setitem__(
+        self,
+        inds: Union[int, slice, Tuple[int, int], Tuple[slice, slice]],
+        view: FigureLike,
+    ):
+        new_view = view.copy(ax=self.fig.add_subplot(self.gs[inds]))
```
YooSunYoung, line 82: How about calling `copy_figure` directly here instead of using `view.copy`?
Then we don't need to have `copy` method in all other backend figure types.

src/plopp/backends/matplotlib/static.py
```diff
@@ -15,29 +15,44 @@ def _make_svg_repr(fig):
...
 class StaticFig:
     """
     Create a static figure to represent one-dimensional data.
     """
 
     def __init__(self, *args, FigConstructor, **kwargs):
         self._fig = FigConstructor(*args, **kwargs)
+        self._args = args
+        self._kwargs = kwargs
+        self._fig_constructor = FigConstructor
```
YooSunYoung, line 37: If this part is going to be repeated in other backend figure wrappers,
maybe we can extract them like this...?

```
class FigureWrapper:
    def __init_figure__(self, *args, FigConstructor, **kwargs):
        self._fig = FigConstructor(*args, **kwargs)
        self._args = args
        self._kwargs = kwargs
        self._fig_constructor = FigConstructor

class StaticFig(FigureWrapper):
    def __init__(self, *args, FigConstructor, **kwargs):
        self.__init_figure__(*args, FigConstructor, **kwargs)

class InteractiveFig(FigureWrapper, VBox):
    def __init__(self, *args, FigConstrutor, **kwargs):
        self.__init_figure__(*args, FigConstructor, **kwargs)
        # other VBox stuff
```

src/plopp/backends/matplotlib/static.py
```diff
@@ -15,29 +15,44 @@ def _make_svg_repr(fig):
...
+        form = 'png' if (npoints > 20_000) else 'svg'
+    return REPR_MAP[form.lower()]
+
+
 class StaticFig:
     """
     Create a static figure to represent one-dimensional data.
     """
 
     def __init__(self, *args, FigConstructor, **kwargs):
```
YooSunYoung, line 33: And just out of curiosity, is there a specific reason that `FigConstructor` is placed between `*args` and `**kwargs`...?
If any of `kwargs` for `FigConstructor` has the key called `FigConstructor`, it will not be passed, right...?
Or was this the intentino not to have unexpected nested calls...?

src/plopp/backends/matplotlib/static.py
nvaytet: I had issues in the past with multiple inheritance on `VBox`, but maybe it wasn't using a mixin.
I will try again, and maybe then I can clean up some other parts.

src/plopp/backends/matplotlib/static.py
nvaytet: Not sure how else you would do it?
`FigConstructor` needs to be a mandatory argument, so I thought this is the place to put it?

src/plopp/backends/matplotlib/static.py
nvaytet: It seems to work nicely, and I removed a bunch of duplication in the process :+1:

src/plopp/backends/matplotlib/static.py
YooSunYoung: It could be the first positional argument, I thought.
Because, when you are passing all keywords, this could happen.
```
class ClumpsyFig:
    def __init__(self, *, FigConstructor: bool):
        self._figure_constructor = FigConstructor

class Figure:
    def __init__(self, *args, FigConstructor, **kwargs):
        self._figure = FigConstructor(*args, **kwargs)

wrapped_fig = Figure(FigConstructor=ClumpsyFig) 
# TypeError: __init__() missing 1 required keyword-only argument: 'FigConstructor'
```
But... there is very low chance that `matplotlib` or other plotting libraries
to add an argument with the name `FigConstructor`, so maybe it was too much of worry ...?

Another good thing about having the `FigConstructor` as the first positional argument might be
that it will have a similar syntax as `functools.partial`
, which means it passes `*args` and `**kwargs`` to the wrapped callable object.

nvaytet: It would however mean that when I use it, I have to do
`return backends.figure1d(FigLine, *args, **kwargs)`
instead of
`return backends.figure1d(*args, FigConstructor=FigLine, **kwargs)`
which may be slightly less readable?

nvaytet: Or maybe it's more readable, I don't know anymore...

YooSunYoung: Sorry I think I should have asked in a separate issue : D....

PR #230, opened 2023-08-02
Remove crop

### LINKED ISSUE #182: Remove the `crop` argument?
Currently, the documented way of setting axis limits is via the `crop` argument, and the syntax is pretty ugly (see also #180).
For example, from the docs:
```Py
da.plot(crop={'x': {'min': sc.scalar(20, unit='m'),
                     'max': sc.scalar(35, unit='m')},
              'y': {'min': sc.scalar(10, unit='m')}})
```
This is rather non-standard (most libraries have something like `xlim/ylim` or `xrange/yrange`), and it mostly arises from [Scipp-ADR0007](https://scipp.github.io/reference/developer/adr/0007-do-not-support-args-referencing-x-or-y.html).

It also adds code complexity in a number of places (such as [here](https://github.com/scipp/plopp/blob/main/src/plopp/functions/common.py#L123)).

A simple way of setting axis limits is to slice your data before you plot, e.g.
```Py
da['x', 1:10]['y': :20].plot()
```
This however does not set the exact axis limits, in the case of value-based slicing with bin-edge coordinates, or for 1d plots where padding is always added to the left and right of the plotted points.

`crop` was introduced to perform both the slicing of the data (so that we don't first plot all the data and then set the axis limits, thus reducing cost) and then setting the exact axes limits.
In practise, this has so far been relatively unused.

There is another (undocumented) way to set the axes limits, which is similar to what Matplotlib would do, which is to first plot the data, and then set the exact range. This is done via the `xrange` and `yrange` properties of the canvas.
```Py
p = pp.plot(pp.data.data2d())
p.canvas.xrange = (10, 35)
p
```

We could just use this mechanism and drop the `crop` argument?
This however does the same as Matplotlib, in the sense that it will first plot the full data and then set axis limits. Note that Matplotlib **does** optimize this if the limits are set before the figure is actually rendered in the notebook.

The downside is that it goes against [ADR-0007](https://scipp.github.io/reference/developer/adr/0007-do-not-support-args-referencing-x-or-y.html).

nvaytet:
Fix #182


YooSunYoung (approved):

nvaytet:
Closing in favour of #234 due to too many merge conflicts

PR #231, opened 2023-08-02
Internal refactor of figures and views

nvaytet:
Internally, it has gotten quite confusing what is a `Figure` and what is a `View`.
Figure currently refers to multiple things:
- the object returned by the `plot` function (and similar) which contains another "figure" inside, and maybe a toolbar also
- the "figure" mentioned above which could be a `FigLine` or a `FigImage`, which contains a canvas
- the underlying matplotlib figure, contained in the `FigLine`'s canvas.

This is getting difficult to navigate.

This PR proposes to move things around and rename for (hopefully) improved clarity:
- the `FigLine` and `FigImage` and now called views: `LineView` & `ImageView`
- the object returned by the `plot` function is still a figure. It inherits methods from the common `Figure` mixin, which in turn inherits from the `BaseFig` (now just a mixin class)
- the underlying matplotlib figure is still called `fig`


src/plopp/backends/matplotlib/utils.py
```diff
@@ -98,7 +98,7 @@ def is_sphinx_build() -> bool:
 
 def copy_figure(fig: FigureLike, **kwargs) -> FigureLike:
     out = fig.__class__(
-        fig._fig_constructor,
+        fig._view.__class__,
```
YooSunYoung, line 101: In case we are not keeping the callable anymore,
should we make sure if the `_view.__class__` is the `View` in `def __init_figure__(self, View, *args, **kwargs):`...?

For example,
```
class FigBase:
  def __init_figure__(self, View, *args, **kwargs):
    self._view = View(*args, **kwargs)
    if not self._view.__class__ is View:
        raise TypeError
```

Cause, you might have something like
```
@dataclasses
class Figure:
  width: int
  height: int

def create_figure(size) -> Figure:
  return Figure(size[0], size[1])

fig = InteractiveFigure(create_figure, size=(10, 10))
```

In this case above, it will create the `fig` as intended, but `copy_figure` will not work.


Or is it too much of worry...? 

```

YooSunYoung (approved):

PR #232, opened 2023-08-03
Legend control

### LINKED ISSUE #206: Add legend control
We need options in the `plot` function to hide the legend and control its placement.
Would be something like
```Py
pp.plot(dict, legend=False)
pp.plot(dict, legend=(0.1, 0.9))
```
The latter would control the location, see the `loc` parameter in https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.legend.html.

nvaytet:
Fix #206


nvaytet:
Closing in favour of #233

PR #233, opened 2023-08-07
Add legend control to plots

### LINKED ISSUE #206: Add legend control

nvaytet:
Fix #206

### DIFF
--- a/src/plopp/backends/matplotlib/line.py
+++ b/src/plopp/backends/matplotlib/line.py
@@ -163,6 +163,11 @@ class Line:
             leg_args = {}
             if isinstance(self._canvas._legend, (list, tuple)):
                 leg_args = {'loc': self._canvas._legend}
+            elif not isinstance(self._canvas._legend, bool):
+                raise TypeError(
+                    "Legend must be a bool, tuple, or a list, "
+                    f"not {type(self._canvas._legend)}"
+                )
             self._ax.legend(**leg_args)
 
     def _make_data(self) -> dict:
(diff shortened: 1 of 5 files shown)


src/plopp/backends/matplotlib/line.py
```diff
@@ -159,8 +159,11 @@ def _make_line(
                 fmt="none",
             )
 
-        if self.label:
-            self._ax.legend()
+        if self.label and self._canvas._legend:
+            leg_args = {}
+            if isinstance(self._canvas._legend, (list, tuple)):
+                leg_args = {'loc': self._canvas._legend}
+            self._ax.legend(**leg_args)
```
YooSunYoung, line 166: Should we throw a `TypeError` if the `legend` argument is not `bool` or `tuple`...?
Or is it not really worth adding...?

src/plopp/backends/matplotlib/line.py
```diff
@@ -159,8 +159,11 @@ def _make_line(
                 fmt="none",
             )
 
-        if self.label:
-            self._ax.legend()
+        if self.label and self._canvas._legend:
```
YooSunYoung, line 162: For navigating users, how about throwing a warning if `label` is not set but `_legend` is `True`...?
i.e. With the instruction how to set `name` attribute to the data array ...?

YooSunYoung (approved):

PR #239, opened 2023-08-21
Use trusted publishers for Pypi package uploads

nvaytet:


.github/workflows/release.yml
```diff
@@ -66,21 +66,22 @@ jobs:
     name: Deploy packages
     needs: [build_conda, build_wheels]
     runs-on: ubuntu-22.04
+    permissions:
+      id-token: write
```
jl-wynen, line 71: This should be used at as small a scope as possible. So since we don't use it for conda (yet?) we should split this into two jobs, I think.

jl-wynen (approved):

SimonHeybrock (approved):

PR #241, opened 2023-08-30
Remove attrs use

SimonHeybrock:
Prepare for deprecation of `DataArray.attrs`.

The previous use of `attrs` that `plopp` cared about has anyway been changed previously: Unaligned coordinates are already stored in `coords` with unset alignment flag since Scipp 23.07. Therefore, I expect that this change has little impact.

### DIFF
--- a/src/plopp/backends/matplotlib/image.py
+++ b/src/plopp/backends/matplotlib/image.py
@@ -99,12 +99,12 @@ class Image:
         for i, k in enumerate('yx'):
             to_dim_search[k] = {
                 'dim': self._data.dims[i],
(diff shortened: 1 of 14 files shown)


nvaytet (approved):

nvaytet:
This will mean that old code that uses attrs would potentially not work anymore.
Do we just then rely on users installing an older version of Plopp to get this to work?

SimonHeybrock:
> This will mean that old code that uses attrs would potentially not work anymore. Do we just then rely on users installing an older version of Plopp to get this to work?

If you refer to Scipp older than 23.07, then yes. If they have at least 23.07 it should work.

PR #245, opened 2023-09-05
Small fix in autoscaling limits with log scale

nvaytet:
Raise error when there are no positive values in limits search and scale is log.

### DIFF
--- a/tests/core/limits_test.py
+++ b/tests/core/limits_test.py
@@ -66,10 +66,13 @@ def test_find_limits_all_zeros():
     assert sc.identical(lims[1], sc.scalar(0.0, unit='s'))
 
 
-def test_find_limits_all_zeros_log_raises():
+def test_find_limits_all_zeros_log_uses_default_positive_values():
     x = sc.zeros(sizes={'x': 5}, unit='s')
-    with pytest.raises(ValueError, match="No positive values were found in array"):
-        _ = find_limits(x, scale='log')
+    lims = find_limits(x, scale='log')
+    zero = sc.scalar(0.0, unit='s')
+    assert lims[0] > zero
+    assert lims[1] > zero
+    assert lims[0] < lims[1]
 
 
 def test_fix_empty_range():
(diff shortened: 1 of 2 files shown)


tests/core/limits_test.py
```diff
@@ -59,6 +59,19 @@ def test_find_limits_no_finite_values_raises():
...
+    x = sc.zeros(sizes={'x': 5}, unit='s')
+    lims = find_limits(x)
+    assert sc.identical(lims[0], sc.scalar(0.0, unit='s'))
+    assert sc.identical(lims[1], sc.scalar(0.0, unit='s'))
+
+
+def test_find_limits_all_zeros_log_raises():
+    x = sc.zeros(sizes={'x': 5}, unit='s')
+    with pytest.raises(ValueError, match="No positive values were found in array"):
+        _ = find_limits(x, scale='log')
```
SimonHeybrock, line 72: What is the effect of this change on an actual plot? What happens when toggling the 'log' button on such a plot?

nvaytet: Yeah it's a good point. The first version of the fix I wrote just set the limits to `[0.1, 1]` instead of raising an error.
Then, when writing the docstring, I thought it should be consistent with the other case where there are no finite values in the input, so I changed to raise an error.

But you are right that if you have a plot with just zeros, and then you click 'log' then it doesn't update the plot, but just raises an error instead. So I've gone back to `[0.1, 1]`

SimonHeybrock: Would it be an option to disable the button?

nvaytet: That would connect too many things together I feel.

What do you do if you have an interactive plot with a slider, and on the first slice, you can see data, and you have set `norm='log'`.
Then you move to the second slice where all values are zero. You can't disable the button because you are already in log, and the plot will just raise an error and the position of the slider will be out if sync with the curve displayed on the canvas becase it won't have updated and will still show the first slice.

I think the default limits is not the cleanest but it's what will give the less headaches in practise. It is also simple, and does not require communication between the canvas and the toolbar, when new values are sent to update the curve.

SimonHeybrock (approved):

PR #249, opened 2023-09-12
Fix axis label for transposed 2d coord

### LINKED ISSUE #246: Axis mislabeled with 2-D coord that is "transposed"
Example:
```python
import scipp as sc
import plopp as pp

a = sc.linspace('a', 0,1,10, unit='m')
b = sc.linspace('b', 0,2,5, unit='s')

da = sc.DataArray(a*b, coords={'a':a, 'b':b*a})
da.plot()
```
Yields
image

Note that the unit is correct, but apparently Plopp uses the name of the inner dim of a coord to get the axis name.

nvaytet:
Fix #246

### DIFF
--- a/tests/graphics/imageview_test.py
+++ b/tests/graphics/imageview_test.py
@@ -136,3 +136,6 @@ def test_axis_label_with_transposed_2d_coord():
     da = sc.DataArray(a * b, coords={'a': a, 'b': b * a})
     fig = ImageView(Node(da))
     assert fig.canvas.xlabel == 'b [m*s]'
+    da = sc.DataArray(a * b, coords={'a': a, 'b': a * b})
+    fig2 = ImageView(Node(da))
+    assert fig2.canvas.xlabel == fig.canvas.xlabel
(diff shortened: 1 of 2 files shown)


tests/graphics/imageview_test.py
```diff
@@ -128,3 +128,11 @@ def test_colorbar_label_has_no_name_with_multiple_artists():
...
     fig = ImageView(Node(a), Node(b))
     assert fig.canvas.cblabel == '[K]'
+
+
+def test_axis_label_with_transposed_2d_coord():
+    a = sc.linspace('a', 0, 1, 10, unit='m')
+    b = sc.linspace('b', 0, 2, 5, unit='s')
+    da = sc.DataArray(a * b, coords={'a': a, 'b': b * a})
+    fig = ImageView(Node(da))
+    assert fig.canvas.xlabel == 'b [m*s]'
```
SimonHeybrock, line 138: Maybe worth testing that we get the same with `'b': a * b`?

SimonHeybrock (approved):

PR #250, opened 2023-09-12
Catch unsupported dtypes when preprocessing input data

### LINKED ISSUE #240: Plotting data with `vector3` dtype plots components
This was unexpected:

```python
import scipp as sc

sc.vectors(dims=['vec'], values=[[1,2,3],[4,5,6]], unit='m').plot()
```
image

I would have expected an error.

nvaytet:
Fixes #240

### DIFF
--- a/src/plopp/plotting/common.py
+++ b/src/plopp/plotting/common.py
@@ -75,28 +75,39 @@ def _check_size(da: sc.DataArray):
         )
 
 
-def check_not_binned(obj):
-    if obj.bins is not None:
-        params = ', '.join([f'{dim}=100' for dim in obj.dims])
+def check_not_binned(da: sc.DataArray):
+    """
+    Plopp cannot plot binned data.
+    This function will raise an error if the input data is binned.
+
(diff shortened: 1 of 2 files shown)


src/plopp/plotting/common.py
```diff
@@ -86,6 +86,20 @@ def check_not_binned(obj):
...
+        sc.DType.affine_transform3,
+        sc.DType.linear_transform3,
+        sc.DType.rotation3,
+        sc.DType.translation3,
+        sc.DType.vector3,
+    )
+    if obj.dtype in forbidden:
+        raise TypeError(
+            f'The input has dtype {obj.dtype} which is not supported by Plopp.'
+        )
```
SimonHeybrock, line 100: You could make this more proof for the future by using something like `obj.shape == obj.values.shape`, maybe?

nvaytet: That is a nice suggestion :+1:

SimonHeybrock (approved):

PR #251, opened 2023-09-12
Fix setting errorbar color for lines in MPL backend

### LINKED ISSUE #235: Changing line color in superplot via widget does not update the color of the error bars
Screenshot at 2023-08-08 15-53-44

nvaytet:
Fixes #235

### DIFF
--- a/src/plopp/backends/matplotlib/line.py
+++ b/src/plopp/backends/matplotlib/line.py
@@ -257,5 +257,6 @@ class Line:
     def color(self, val):
         self._line.set_color(val)
         if self._error is not None:
-            self._error.set_color(val)
+            for artist in self._error.get_children():
+                artist.set_color(val)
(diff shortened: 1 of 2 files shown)


jokasimr:
Looks good to me!

One question though, when I create an errorbar using `plt.errorbar(x, y, s, fmt='none')`  and try to access the hex value of the color using `to_hex(eb.get_children()[0].get_color())` I get the error `ValueError: RGBA sequence should have length 3 or 4`.

I haven't been able to clone the plopp project and run the tests myself, but I guess that doesn't happen there, what's the difference?

jokasimr (approved):
Looks good to me.

jokasimr:
Could be, I'm on `3.3.2`.

nvaytet:
I think we require `>=3.5` so maybe we are ok?

PR #254, opened 2023-09-20
Make high-level plotting functions accept nodes as inputs

### LINKED ISSUE #247: Generalizing plotting functions to accept nodes as input
Right now, we have a lot a functionality provided by the functions in the `plotting` submodule.
However, they will only accept raw data as input.
A much improved experience for someone trying to build their own small interactive visualization would be if those functions could be used as part of a larger graph of nodes.

For example, say a user has some 3d data, and they build an interface where a couple of operations are happening on the data, e.g. smoothing, and the parameters are controlled via widgets. Then, at the end of the graph, the user would like to view the data as a 2d image with a slider. Currently, they have to go in the source code for the `slicer` plot and copy how the slider is hooked up to the figure. It would be much better if the `slicer` would also accept nodes as input and could thus be directly connected to the graph.

In principle, this should be relatively simple to implement. It will mostly involve checking types on input and wrapping raw data into a node if needed.

In addition, I feel that all the logic in those high-level functions (e.g. the [preprocess](https://github.com/scipp/plopp/blob/main/src/plopp/plotting/common.py#L93) function) should be put into nodes that would then be visible steps in a `show_graph` visualization.

See https://github.com/scipp/plopp/tree/more-nodes for a draft of how this may look like,

nvaytet:
Fixes #247


src/plopp/plotting/common.py
```diff
@@ -175,18 +176,21 @@ def preprocess(
...
-    else:
-        return [preprocess(obj, **kwargs)]
+    inputs:
+        The inputs to be converted to nodes.
+    processor:
+        The function that will be applied to each input to convert it to a node.
+    """
+    flat_inputs = []
+    for inp in inputs:
+        if hasattr(inp, 'items') and (not hasattr(inp, 'between')):
```
jl-wynen, line 192: Is this the thing you asked ChatGPT about? Have you considered using something like
```python
def is_pandas_series(x):
    try:
        import pandas as pd
    except ImportError:
        return False
    return isinstance(x, pd.Series)
```
The current check seems too brittle

tests/plotting/inspector_test.py
```diff
@@ -6,6 +6,49 @@
...
+    fig2d.toolbar['inspect']._tool.remove(0)
+    assert len(fig1d.artists) == 0
+
+
+def test_operation(use_ipympl):
+    da = pp.data.data3d()
+    ip_sum = pp.inspector(da, operation='sum')
+    ip_mean = pp.inspector(da, operation='mean')
+    assert ip_sum[0][0]._view.colormapper.vmax > ip_mean[0][0]._view.colormapper.vmax
+    assert ip_sum[0][0]._view.colormapper.vmin < ip_mean[0][0]._view.colormapper.vmin
```
jl-wynen, line 49: These tests don't seem to be related to the purpose of this PR?

nvaytet: Yeah they are not, I should move them to another PR.

src/plopp/plotting/common.py
nvaytet: Ah, so if it is a pandas series, then it means you **must** have pandas installed... I guess that should work :+1:

src/plopp/plotting/inspector.py
```diff
@@ -70,31 +79,27 @@ def inspector(
...
-    if obj.ndim != 3:
-        raise ValueError(
-            'The inspector plot currently only work with '
-            f'three-dimensional data, found {obj.ndim} dims.'
-        )
     require_interactive_backend('inspector')
 
-    da = preprocess(obj, ignore_size=True)
-    in_node = Node(da)
+    in_node = Node(preprocess, obj, ignore_size=True)
```
jl-wynen, line 84: What if `obj` is a node?

src/plopp/core/typing.py
```diff
@@ -11,3 +18,9 @@ class VisibleDeprecationWarning(UserWarning):
 
 
 VisibleDeprecationWarning.__module__ = 'plopp'
+
+Plottable = Union[VariableLike, ndarray, Node]
+
+PlottableMulti = Union[
+    VariableLike, ndarray, Dict[str, Union[VariableLike, ndarray]], Node
```
jl-wynen, line 25:
```suggestion
    VariableLike, ndarray, Dict[str, Union[VariableLike, ndarray, Node]], Node
```
?

nvaytet: Can we just have `PlottableMulti = Union[Plottable, Dict[str, Plottable]]`?

jl-wynen: Yes, good point!

src/plopp/plotting/inspector.py
nvaytet: it just stays a Node if it is already a Node: https://github.com/scipp/plopp/blob/main/src/plopp/core/node.py#L70

jl-wynen (approved):

PR #256, opened 2023-09-21
fix: filter largest and smallest values of dtypes

jokasimr:
This change makes the matplotlib backend ignore the largest and smallest values of the datatype of the variables plotted.

The rationale for this behavior is that the largest and smallest values are almost only reached when a max- or min-operation is performed on an empty bin.

**Example code**
```python
import scipp as sc

filename = sc.data.rhessi_flares()
flares = sc.io.load_hdf5(filename)
flares_ds = sc.Dataset(
    coords=flares.coords,
    data=dict(
        flares=flares.data,
        max_energy=flares.attrs['max_energy'],
    )
)
flares_ds['max_energy'].bin(peak_time=140).bins.max().plot()
```
Result before this change:
Screenshot from 2023-09-21 16-00-57
Result after this change:
Screenshot from 2023-09-21 16-03-12

Some questions:

* I'm not actually sure this is the best option. Is it better to let the user do this filtering?
* Does plopp support signed integer dtypes? If so we will have to deal with that case separately.
* Should this be implemented at a higher level, e.g. not in the matplotlib backend?

### DIFF
--- a/src/plopp/backends/matplotlib/line.py
+++ b/src/plopp/backends/matplotlib/line.py
@@ -24,6 +24,18 @@ def _parse_dicts_in_kwargs(kwargs, name):
     return out
 
 
+def _nan_if_not_finite(x: np.array):
+    if not hasattr(x, 'dtype'):
+        return x
+    if np.issubdtype(x.dtype, np.integer):
+        info = np.iinfo(x.dtype)
+    elif np.issubdtype(x.dtype, np.floating):
+        info = np.finfo(x.dtype)
+    else:
+        return x
+    return np.where(np.isfinite(x) & (x != info.max) & (x != info.min), x, np.nan)
+
+
 class Line:
     """
     Artist to represent one-dimensional data.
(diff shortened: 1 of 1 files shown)


src/plopp/backends/matplotlib/line.py
```diff
@@ -193,12 +205,23 @@ def _make_data(self) -> dict:
             y = sc.concat([y[0:1], y], dim=self._dim)
             if mask is not None:
                 mask['y'] = np.concatenate([mask['y'][0:1], mask['y']])
-        return {
-            'values': {'x': x.values, 'y': y.values},
-            'stddevs': error,
-            'mask': mask,
-            'hist': hist,
-        }
+        return self._set_to_nan_if_not_finite(
```
nvaytet, line 208: I am wondering if we actually want to do this.
Say for example this was already in plopp. Would you have noticed you had strange values in your data? Probably not. So maybe this is a good thing?

Or maybe it could be activated from a keyword arg? But then that arg would have to be propagated all the way to the top...

jokasimr: Yeah that is an important point. I think this feature is too obscure to warrant a keyword argument.

nvaytet: After discussion, the conclusion is that we do not make this change, but we can revisit if users start complaining or asking for this fix.

Also note that we need to very very careful about performing any modifications on the data that is given as input. We've had [issues in the past](https://scipp.github.io/reference/developer/adr/0013-plot-should-not-resample-automatically.html) with this, and we concluded that anything that is plotted needs to be an accurate representation of the data that is given to us.

PR #257, opened 2023-09-22
Use `[datetime]` instead of unit on datetime axes

### LINKED ISSUE #255: Wrong unit on datetime axis
```python
import scipp as sc

sc.DataArray(
    coords={
        'x': sc.arange(
            'x',
            start=sc.datetime(0, unit='s'),
            stop=sc.datetime(int(1e9), unit='s'),
            step=sc.scalar(int(2e7), unit='s'),
        ),
    },
    data=sc.ones(dims=['x'], shape=[50])
).plot()
```
Screenshot from 2023-09-21 13-11-11

The numbers on the x-axis correspond to years, but the displayed unit is "[s]".

nvaytet:
Fix #255


nvaytet:
I'm not sure this is the right choice. A plot with a datetime coordinate now looks the same as a coordinate with `unit=None`...

YooSunYoung:
Can it be rather handled specifically, i.e. by investigating ``ticks`` of the plot...?

nvaytet:
> Can it be rather handled specifically, i.e. by investigating ticks of the plot...?

I guess when you have a wide time range showing years, e.g. `[2003, 2004, 2005, 2006]`, and the coordinate name is something like `time`, then it would make sense to have `time [year]` as the unit.

However, if you zoom in:

1. you would have to change the unit depending on the level of zoom, which makes things complicated and potentially expensive (see [this old notebook](https://github.com/scipp/scipp/blob/de0195e89dc7514d4ec855ae6104b9407b2fc01f/docs/visualization/plotting/datetime_handling.ipynb) which lists all the cases we had to account for)
2. If you have times along the axis, in the format of `12:37:55`, I feel having `time [s]` would be slightly confusing because seconds is only for the last two digits?

YooSunYoung (approved):

PR #259, opened 2023-09-22
docs: mention changing labels by setting `.canvas.ylabel`

jokasimr:
Gives an example of how to change the axes labels by setting the `.canvas.ylabel` property.


docs/customization/tweaking-figures.ipynb
```diff
@@ -77,7 +77,7 @@
    "id": "091f34e3-49e2-492e-b70f-f34161e1cb24",
    "metadata": {},
    "source": [
-    "The location of the ticks, as well as their labels, can also be changed in the same way"
+    "To change the location of the ticks, as well as their labels, we act directly on the Matplotlib axes via the `.ax` property of the figure:"
```
nvaytet, line 80: Since the method you are giving now is not matplotlib-specific, I think I would restucture the notebook a little.
We could first have a section which says nothing about matplotlib, that would include changing axes labels and setting figure title.

We could then have a section on 'further modifications' or something like that, that would include changing tick labels as well as the stuff on controlling axes placement?
We would also move the sentence from the intro that begins with "Instead of providing keyword arguments for tweaking every aspect of the figures..." into that section.

docs/customization/tweaking-figures.ipynb
```diff
@@ -56,46 +76,75 @@
...
+   "metadata": {
+    "editable": true,
+    "slideshow": {
+     "slide_type": ""
+    },
+    "tags": []
+   },
    "source": [
-    "The location of the ticks, as well as their labels, can also be changed in the same way"
+    "There are also other properties we can set on the canvas object such as the axis range:"
```
nvaytet, line 118: Maybe make a new subsection like `## Setting the displayed range`?

docs/customization/tweaking-figures.ipynb
```diff
@@ -106,12 +155,68 @@
...
+   "id": "93c474b1-fcaa-4c96-b026-1e17ed278e72",
+   "metadata": {
+    "editable": true,
+    "slideshow": {
+     "slide_type": ""
+    },
+    "tags": []
+   },
+   "source": [
+    "# Further modifications\n",
```
nvaytet, line 181: I think having a top-level section here (with a single '#') may mess up the table of contents in Sphinx.

How about having the following sections:
```
# Tweaking figures
  ## Basic customization
    ### Changing axes labels
    ### Setting the axis range
    ### Adding a title
  ## Further modifications
    ### Tick locations and labels
    ### Axes placement
```
?

nvaytet (approved):

PR #262, opened 2023-09-26
Add simple xyplot helper function

nvaytet:
To quickly plot a variable as a function of another, without having to manually make a DataArray from them.

This has been requested by a couple of users.


src/plopp/plotting/xyplot.py
```diff
@@ -0,0 +1,27 @@
...
+        The variable to use as the coordinates for the horizontal axis.
+    y:
+        The variable to use as the data for the vertical axis.
+    **kwargs
+        See :py:func:`plopp.plot`.
+    """
+    x = to_variable(x)
+    y = to_variable(y)
+    dim = x.dim
+    da = sc.DataArray(data=y.rename_dims({y.dim: dim}), coords={x.dim: x})
```
jl-wynen, line 26: I'm not sure you should rename dims here. This can potentially hide errors when people try to plot unrelated things. I think that, if they want to do that, they should be explicit and rename dimensions themselves.

nvaytet: This would however mean that we can (usually) not plot a Variable as a function of a numpy array, because the numpy array gets converted to a Variable with a `axis-0` dimension.

But maybe that is also ok? Plotting numpy array against numpy array should be fine.

jl-wynen: You could use a special placeholder dimension that would get replaced here. But it might not be worth the complication. And I tend to agree that we shouldn't rename ndarray dims like you say. Because we don't allow mixing variables and ndarrays anywhere else.

jl-wynen (approved):

PR #270, opened 2023-10-02
Fix connecting draw tool to node

nvaytet:
When using the `add_child` method on a `Node`, this only appended to the list of children, but did not automatically update the list of parents/kwparents.

So we remove the `add_child` because it is somewhat ambiguous if the parent would then be added to the `parents` or `kwparents`. Instead, we replace it with two `add_parents` and `add_kwparents`.

We also fix the draw tool when the destination is a node and not a figure.

### DIFF
--- a/src/plopp/core/node.py
+++ b/src/plopp/core/node.py
@@ -5,9 +5,19 @@ from __future__ import annotations
 
 import uuid
 from itertools import chain
-from typing import Any, Union
+from typing import Any, List, Union
 
 from .view import View
+
+
+def _no_replace_append(container: List[Node], item: Node, kind: str):
+    """
+    Append ``item`` to ``container`` if it is not already in it.
+    """
+    if item in container:
+        tpe = 'View' if kind == 'view' else 'Node'
+        raise ValueError(f"{tpe} {item} is already a {kind} in {container}.")
+    container.append(item)
 
 
 class Node:
@@ -43,8 +53,6 @@ class Node:
         self.kwparents = {
             key: p if isinstance(p, Node) else Node(p) for key, p in kwparents.items()
         }
-        for parent in chain(self.parents, self.kwparents.values()):
-            parent.children.append(self)
         self._data = None
 
         if func_is_callable:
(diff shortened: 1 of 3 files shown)


src/plopp/core/node.py
```diff
@@ -120,11 +120,21 @@ def request_data(self) -> Any:
...
-    def add_child(self, child: Node):
+    def add_parents(self, *parents: Node):
         """
-        Add a child to the node.
+        Add one or more parents to the node.
         """
-        self.children.append(child)
+        for parent in parents:
+            self.parents.append(parent)
+            parent.children.append(self)
```
jl-wynen, line 129: Is there a risk to add the same parent multiple times? Should the `parents` and `children` attributes be sets instead of lists?

nvaytet: Maybe I should do that yes. I will try to see how this looks like in tests.

nvaytet: Do you think users may want to control the order in which updates are made? Because we would maybe lose that ability by using sets?
My feeling is that building something that relies on an order sounds flaky and should be avoided anyway, but maybe there are cases where avoiding this is impossible?

nvaytet: So I don't think we can use sets, at least not for the `parents`, because they represent the arguments to the function and they should preserve order.

jl-wynen: So it would also be ok to have the same parent multiple times if you want its output to be passed as multiple function args? Then only the children need to be unique to prevent notifying them multiple times, I guess.

nvaytet: I think having the same parent twice would indicate you have messed up somewhere, since in principle there is no reason to have the same value twice as different function arguments?

Unless you are thinking of when writing generic code, you could send the same data as e.g. `signal` and `background` and subtract one from the other?

jl-wynen: But it would probably result in some strange/useless output in most cases.

jl-wynen (approved):

PR #275, opened 2023-11-03
Add residual plot overloading the `-` operator

### LINKED ISSUE #201: Add a new `residuals_plot`?
[This example](https://scipp.github.io/plopp/customization/tweaking-figures.html#Axes-placement) shows how to make a plot of multiple lines with a residual difference panel below the main plot.

Is this something that is common enough that we could have a dedicated function for creating it?
For example, `plopp.residual_plot(b, c, reference=a)` would show 3 lines in the main panel and `b-a` and `c-a` in the residuals panel.

I guess it would be most useful if it could be used as a view in a graph of nodes, and not just a single purpose function.
This would mean that it could be used in combination with a slider, where both the main panel and the residuals would update when moving the slider.

nvaytet:
Note that this was just to try things out.
Decision on whether we should implement it this way has not yet been made.

Fixes #201


jl-wynen:
Sorry for the late reply. This got lost in my feed somehow.

I don't think using the `-` operator like this is a good idea. The other operators only compose plots. But here, it adds a new line to an existing plot and computes new data for the residuals plot.

This solution optimises completely for modularity but, as you said, it's expensive because of the temporary plots. Would it be acceptable to modify the input plot by adding a line to it? This would give more or less the same performance as a dedicated function as first described in #201 but would also be modular. (In this case, `-` would definitely be a bad choice because it doesn't look like it modifies the input.)

nvaytet:
> Sorry for the late reply. This got lost in my feed somehow.
> 
> I don't think using the `-` operator like this is a good idea. The other operators only compose plots. But here, it adds a new line to an existing plot and computes new data for the residuals plot.

This isn't technically modifying the original plot, it makes a new plot and copies the data over from the old plots, which should still be displayable in their original form.

> This solution optimises completely for modularity but, as you said, it's expensive because of the temporary plots. Would it be acceptable to modify the input plot by adding a line to it? This would give more or less the same performance as a dedicated function as first described in #201 but would also be modular. (In this case, `-` would definitely be a bad choice because it doesn't look like it modifies the input.)

Maybe this isn't so expensive anymore after #290 ?

jl-wynen:
> This isn't technically modifying the original plot, it makes a new plot and copies the data over from the old plots, which should still be displayable in their original form.

True. But it does compute new data which the other operators don't do.

> Maybe this isn't so expensive anymore after #290 ?

Do you have benchmarks?

nvaytet:
> Do you have benchmarks?

Nothing beyond what I mention in the PR message:

- Creating the 20 subplots now takes a little over 1s as opposed to 1 minute before.
- As a bonus, running the unit tests now takes 30s instead of a minute

src/plopp/backends/matplotlib/residuals.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    Create a residual plot from two figures, using the data from the second figure as
+    the reference the residuals are computed from.
+
+    Parameters
+    ----------
+    main_fig:
+        The main figure.
+    reference:
+        The reference figure.
```
jl-wynen, line 36: Missed this earlier, but wouldn't it make more sense to accept a data array instead of a figure? Why would you require the user to plot the reference data first?

src/plopp/backends/matplotlib/residuals.py
```diff
@@ -0,0 +1,106 @@
...
+    res_ax = fig.add_subplot(gs[1], sharex=main_ax)
+    main_panel = copy_figure(main_fig, ax=main_ax)
+    main_canvas = main_panel.canvas
+    if main_canvas.is_widget():
+        fig.canvas.toolbar_visible = False
+        fig.canvas.header_visible = False
+    ref_node = next(iter(reference.graph_nodes.values()))
+    data = ref_node()
+    if not data.name:
+        data.name = "reference"
```
jl-wynen, line 74: This modifies the input, right? Seems like you should make a copy of `reference` first.

src/plopp/backends/matplotlib/residuals.py
nvaytet: Yes, I'm starting to think that the whole approach using operators is a bad idea. I also ran into problems if, for example, the user made a figure where they changed the line style or the line color. Then I had to make sure that was copied over.

I think we need to go back to some of the suggestions in #201 and use a dedicated function, and drop the operators.

PR #282, opened 2023-11-22
Remove warning about tight_layout when using tiled

nvaytet:


jl-wynen (approved):
Didn't know this is a thing :+1:

PR #283, opened 2023-11-24
Use card previews in docs basics index

nvaytet:
Before:
Screenshot at 2023-11-24 14-58-52

After:
Screenshot at 2023-11-24 14-58-30


docs/basics/index.md
```diff
@@ -1,8 +1,52 @@
...
+
+:::
+
+:::{grid-item-card} Scatter 3D plot
+:link: scatter3d-plot.ipynb
+:img-bottom: ../_static/basics/scatter3d-plot.png
+
+:::
+
+::::
```
jl-wynen, line 45: Looks like a link to `saving-figures` is missing.

jl-wynen (commented):
Nice idea!

docs/basics/index.md
nvaytet: Yes, I didn't add it because I didn't have a good image for it.
As it's still in the left hand side menu, I thought it would be ok without it?

Also, if I want to have a card the same size as the other, I have to add two dummy/empty cards next to it.
I could of course just add a single card that spans the whole width?

nvaytet: Would look like this
Screenshot at 2023-11-24 15-40-19

jl-wynen: Would the dummy cards be visible?

nvaytet: Actually I can force the number of columns
Screenshot at 2023-11-24 16-07-00

jl-wynen: Ok :+1:

jl-wynen (approved):

PR #286, opened 2023-12-13
Random samples histograms in gallery.

YooSunYoung:
I couldn't find any interesting datasets so I just made a table of random samples that you can generate with ``numpy.random`` module.

It was very easy thanks to ``tiled`` feature...!

Here is the all samples plot: 

output


nvaytet (commented):
Can you remove the output from the notebook?

docs/gallery/index.ipynb
```diff
@@ -33,7 +33,8 @@
     "* [3-D scatter plot with threshold](scatter3d-with-threshold.ipynb)\n",
     "* [3-D scatter plot with slider](scatter3d-with-slider.ipynb)\n",
     "* [Interactive masking](interactive-masking.ipynb)\n",
-    "* [Peeling off the layers](peeling-layers.ipynb)"
+    "* [Peeling off the layers](peeling-layers.ipynb)\n",
+    "* [Titled random sample histograms](tiled-random-samples.ipynb)"
```
nvaytet, line 37:
```suggestion
    "* [Tiled random sample histograms](tiled-random-samples.ipynb)"
```

docs/gallery/tiled-random-samples.ipynb
```diff
@@ -0,0 +1,155 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Continuous samples\n",
+    "continuous_samples = plots['normal'] + plots['laplace'] + plots['vonmises']\n",
+    "continuous_samples.fig.set_size_inches((15, 3))\n",
```
nvaytet, line 117: I wonder if this (and the one below) is needed for the docs? I think the figure might get scaled to the page width automatically? You could then reduce the amount of code by just having `plots['binomial'] + plots['logseries']` in the cell?

docs/gallery/index.ipynb
```diff
@@ -33,7 +33,8 @@
...
    ]
```
nvaytet, line 38: I find it is a shame that it's the last figure which appears as the preview in the gallery page:
Screenshot at 2023-12-14 09-39-03

It would look more 'impressive' if it was the first figure. Can you save the first figure as a thumbnail (in a hidden cell at the end of the notebook maybe?) and add it to the `docs/conf.py` (see other notebooks in the gallery for an example of how this is done).

docs/gallery/tiled-random-samples.ipynb
YooSunYoung: That is true... it was only annoying in my notebook....! I'll remove them.

docs/gallery/tiled-random-samples.ipynb
```diff
@@ -95,6 +95,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# Adjust some of plots\n",
+    "plots['uniform'].ax.set_ylim(0, 0.11)\n",
+    "plots['beta'].ax.set_ylim(0, 0.2)\n",
+    "plots['triangular'].ax.set_ylim(0, 0.2)\n",
+    "\n",
```
nvaytet, line 102: At some point we should look at the performance of `tiled` because making this figure is super slow :-( 
It was designed for making a small number of subplots, but you're right that this is probably the first thing people will try to make with it.

nvaytet (commented):
For the thumbnail, look at the last cell of https://github.com/scipp/plopp/blob/main/docs/gallery/masking-a-range.ipynb. It is a hidden cell that saves the figure to the correct place on-the-fly so we don't have to store the thumbnail in the git repo.

docs/gallery/tiled-random-samples.ipynb
YooSunYoung: Yes... it takes about 40 seconds in my machine.
But I thought it was not so important,
cuz you can just use ``matplotlib`` subplots and manually pass ``ax`` to the plotting function if it should be faster.

Do you know why it is so slow...?

nvaytet (approved):

docs/gallery/tiled-random-samples.ipynb
nvaytet: > Do you know why it is so slow...?

I've done a bit of digging. There are a few reasons:

1. Plotting with plopp is just slower (about 3x slower) than pure matplotlib. We do a bunch of preprocessing on the data, converting it to a data array, checking units etc. We also construct more things such as a View, Canvas, Figure, all of which are Plopp-specific objects.

2. The (lazy) way tiled is currently implemented is that every figure is made twice. When you assign a plot to e.g. `tiled[0, 0]`, you first make the plot, and then assign it. When you assign it, there is an operation (`copy_figure`) where a new figure is made from the old one. So you make it twice.

What's kind of strange is that, if I make the subplots manually with matplotlib, and then just iterate over the plots and assign axes to each plot, it takes 30s instead of about 1min using Tiled on my laptop.

But just making the matplotlib subplots and plotting some numpy data and setting axes labels only takes 1s. So there is a factor of 30, not a factor of 3 which is what I saw for just making a single plot.

nvaytet: What's even more strange is that making the same plot in a script and not in jupyter only takes 8 seconds...

nvaytet: I found the culprit: [these lines](https://github.com/scipp/plopp/blob/main/src/plopp/backends/matplotlib/utils.py#L38-L40) are switching to another backend and then switching back when in jupyter. This seems to get slower and slower the larger the figure is (or maybe the more plots there are?).

I cannot remember why I needed this but I am guessing there was a good reason, will try to figure it out.

YooSunYoung: Then I'll just merge it since it's not affecting docs building time : D...!

PR #290, opened 2023-12-20
Improve plot creation performance

nvaytet:
In Jupyter, when creating a matplotlib figure with `plt.figure`, it is then automatically shown below the cell that contains the call to `plt.figure`.
This is undesired behaviour because we instead want the figure to be displayed when the figure repr is requested.

Until now, the workaround was to use a context manager to create the figure
```Py
@contextmanager
def silent_mpl_figure():
    """
    Context manager to prevent automatic generation of figures in Jupyter.
    """
    backend_ = mpl.get_backend()
    revert = False
    if 'inline' in backend_:
        mpl.use("Agg")
        revert = True
    with mpl.pyplot.ioff():
        yield
    if revert:
        mpl.use(backend_)
```
where we would temporarily switch to another backend before switching back.

This turns out to be an expensive operation, partly because when you switch backend, matplotlib closes all the existing figures (through `plt.close()`). And it gets more expensive the more figures you have.

This became very obvious when making a figure with 20 subplots in #286 using `Tiled`.

In this PR, we replace the old context manager with a function that makes a figure using `plt.Figure` instead of `plt.figure()`. This no longer automatically displays the created figure. Finally, when using the interactive `widget` backend, we have to do a bot more work, because the raw `plt.Figure` does not have a toolbar, nor is interactive.

Creating the 20 subplots now takes a little over 1s as opposed to 1 minute before.
As a bonus, running the unit tests now takes 30s instead of a minute :-)

### DIFF
--- a/src/plopp/backends/matplotlib/utils.py
+++ b/src/plopp/backends/matplotlib/utils.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
-from contextlib import contextmanager
 from io import BytesIO
 from typing import Literal
 
@@ -28,20 +27,12 @@ def fig_to_bytes(fig: Figure, form: Literal['png', 'svg'] = 'png') -> bytes:
     return buf.getvalue()
 
 
-@contextmanager
-def silent_mpl_figure():
-    """
-    Context manager to prevent automatic generation of figures in Jupyter.
-    """
-    backend_ = mpl.get_backend()
-    revert = False
-    if 'inline' in backend_:
-        mpl.use("Agg")
-        revert = True
-    with mpl.pyplot.ioff():
-        yield
-    if revert:
-        mpl.use(backend_)
+def make_figure(*args, **kwargs):
+    backend = mpl.get_backend()
+    manager = backend.new_figure_manager(*args, FigureClass=Figure, **kwargs)
+    fig = manager.canvas.figure
+    ax = fig.add_subplot()
+    return fig, ax
 
 
 def is_interactive_backend():
--- a/src/plopp/backends/matplotlib/utils.py
+++ b/src/plopp/backends/matplotlib/utils.py
@@ -5,7 +5,7 @@ from io import BytesIO
 from typing import Literal
 
 import matplotlib as mpl
-from matplotlib.pyplot import Figure
+from matplotlib.pyplot import Figure, _get_backend_mod
 
 from ..protocols import FigureLike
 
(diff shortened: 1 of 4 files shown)


YooSunYoung:
I was trying `interactive-masking` from gallery and the tool seems frozen. Or am I doing something wrong...?

[Screencast from 01-04-2024 10:35:15 AM.webm](https://github.com/scipp/plopp/assets/17974113/b9d4ac07-69f8-4a10-b59c-e3d79abe05bd)

src/plopp/backends/matplotlib/utils.py
```diff
@@ -28,29 +27,35 @@ def fig_to_bytes(fig: Figure, form: Literal['png', 'svg'] = 'png') -> bytes:
...
+    will not have a toolbar nor will it be interactive, as opposed to what
+    ``plt.figure`` returns. We therefore copy the minimal required code inside the
+    ``plt.figure`` function which creates a figure manager (which is apparently what
+    creates the toolbar and makes the figure interactive).
+    """
+    if not is_interactive_backend():
+        return Figure(*args, **kwargs)
+    backend = _get_backend_mod()
+    manager = backend.new_figure_manager(1, *args, FigureClass=Figure, **kwargs)
+    return manager.canvas.figure
```
YooSunYoung, line 56: Maybe ``manager`` should be kept somewhere...?

nvaytet: Would you need it for anything? I thought all we need is the figure...

YooSunYoung: I don't know... 
I was just guessing why [this](https://github.com/scipp/plopp/pull/290#issuecomment-1876784760) happened.

nvaytet:
> I was trying `interactive-masking` from gallery and the tool seems frozen. Or am I doing something wrong...?

No, you are not doing anything wrong. The tool will work if you don't run the cell that contains

```Py
r.value = True

r._tool.click(50, 200)
r._tool.click(200, 250)
r._tool.click(30, 50)
r._tool.click(250, 170)

from ipywidgets import HBox

fig.children = [
    fig.top_bar,
    HBox([fig.left_bar, fig.canvas.to_image(), fig.right_bar]),
    fig.bottom_bar,
]
```
It is needed so that the rectangles actually render on the plot in the docs (the cell is actually hidden in the docs).
It converts the figure to a static image.

Without it, we only see the original image without the masked rectangles.

This is because to make interactive figures, we are inheriting from a `ipywidgets.VBox` object (so you can use figrues inside other layout widgets), and the figure is rendered upon creation instead of when a repr is requested (see [here](https://github.com/scipp/plopp/blob/main/src/plopp/backends/matplotlib/interactive.py#L29) where the `to_widget` will convert to a static image if we are building sphinx docs instead of an actual widget).

I have never been happy with this solution, but matplotlib plots inside widgets don't render in sphinx (you just get a blank plot).
If you have a better suggestion, then I'd be glad to hear it.

YooSunYoung:
Even if I didn't run that cell, the rectangle selection tool still seems frozen while other tools work fine.
But it's not related to this change, cause it happens on the main branch as well... maybe just my workstation...sorry for the unnecessary delay...!

PR #291, opened 2024-01-05
feat: add option to not display masked data

### LINKED ISSUE #266: Allow other ways of handling masks?
Currently, input data that contains masks will still see the masked values plotted on figure.
Masked values are distinguished from other values by either adding a black edge color around the point markers (1d plots), or by using a different colormap (2d plots).

However, sometimes a user really does not want to see the masked values, for example if the masked values are orders of magnitude larger than the part of the data they were interested in.

Example:
```Py
da = pp.data.data1d()
da.values[0] = 1000.
da.masks['large'] = da.data > sc.scalar(1.0, unit=da.unit)
pp.plot(da)
```
Figure 17

The fluctuations in the data cannot be seen because of the spike in the first data point, making the plot rather useless.

Maybe we can simply have an option, e.g. `drop_masked=True` to remove those points?
The question now is: should the default be `drop_masked=True` or `drop_masked=False`?

(Matplotlib has `drop_masked=True` by default for numpy masked arrays).

jokasimr:
Fixes #266.
This still needs some more discussion.

1. Is it better to hide the data by setting it to `nan` instead of manually removing it using the masks? I opted for the manual option to make it distinct from non-masked nan-data.
2. Should this be default or do we keep the old default?
3. In what other plot-types should this option be added?

### DIFF
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -145,19 +145,23 @@ class Canvas:
         draw:
             Make a draw call to the figure if ``True``.
         """
+        xmin = np.inf
+        xmax = np.NINF
+        ymin = np.inf
+        ymax = np.NINF
         if self.ax.lines:
             self.ax.relim()
             self.ax.autoscale()
             xmin, xmax = self.ax.get_xlim()
-            ymin, ymax = self.ax.get_ylim()
+            some_line_had_preferred_ylim = False
             for line in self.ax.lines:
-                if hasattr(line, '_plopp_force_ylim'):
-                    ymin, ymax = line._plopp_force_ylim
-        else:
-            xmin = np.inf
-            xmax = np.NINF
-            ymin = np.inf
-            ymax = np.NINF
+                if hasattr(line, '_plopp_preferred_ylim'):
+                    _ymin, _ymax = line._plopp_preferred_ylim
+                    ymin = min(ymin, _ymin)
+                    ymax = max(ymax, _ymax)
+                    some_line_had_preferred_ylim = True
+            if not some_line_had_preferred_ylim:
+                ymin, ymax = self.ax.get_ylim()
         for c in self.ax.collections:
             if isinstance(c, QuadMesh):
                 coords = c.get_coordinates()
@@ -177,6 +181,14 @@ class Canvas:
                 xmax = max(xmax, right.value)
                 ymin = min(ymin, bottom.value)
                 ymax = max(ymax, top.value)
+        if self.yscale == 'log' and ymin < 0:
+            # Something has gone wrong
+            # probably the y-limit was set before the scale changed
+            # set the y-limit to something reasonable
+            ymin = 1e-2 * ymax
+        if self.xscale == 'log' and xmin < 0:
+            xmin = 1e-2 * xmax
+
         if self._autoscale == 'grow':
             self._xmin = min(self._xmin, xmin)
             self._xmax = max(self._xmax, xmax)
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -150,6 +150,9 @@ class Canvas:
             self.ax.autoscale()
             xmin, xmax = self.ax.get_xlim()
             ymin, ymax = self.ax.get_ylim()
+            for line in self.ax.lines:
+                if hasattr(line, '_plopp_force_ylim'):
+                    ymin, ymax = line._plopp_force_ylim
         else:
             xmin = np.inf
             xmax = np.NINF
--- a/src/plopp/backends/matplotlib/line.py
+++ b/src/plopp/backends/matplotlib/line.py
@@ -9,6 +9,7 @@ import scipp as sc
 from matplotlib.lines import Line2D
 from numpy.typing import ArrayLike
 
+from ...core.limits import find_limits
 from ...core.utils import merge_masks
 from .canvas import Canvas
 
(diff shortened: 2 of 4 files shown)


nvaytet:
My idea for this is that the masked values would always be shown (users found it useful to see what had been masked instead of just having blanks in the curve).
The question was if they should be included when trying to find the limits for the data (vertical axis range for 1d plots and colorbar range for 2D plots).
So I think the code in the line implementation should stay as it is, and instead modify the code that finds the min and max values of the data.

jokasimr:
> My idea for this is that the masked values would always be shown..., and instead modify the code that finds the min and max values of the data.

Aha I see, that makes sense!

jokasimr:
Here are two images displaying how it looks.
Did we decide to make this default or optional?

Screenshot from 2024-01-08 14-19-41

jokasimr:
I have a couple of questions left:

1. Should we have a "padding"? If we use the min/max as lims then some points will be cut in half. Adding a small padding, 5-10% of the figure width prevents this but it might look weird in other cases, for example image plots doesnt benefit from padding. I think matplotlib does this depending on the plot type.
2. When the norm is 'log' and some y-value value is negative then the bounding box strategy we use fails. Preferably we would like the select the bottom limit as the minimum positive y-value, but when we have bounding boxes we can't access the minimum positive value. Right now I just pick a bottom value 0.1 x the top value. Should we do something else instead?
3. Should there be a toggle to turn on/off this new behavior? Should the new behavior be default or optional?

nvaytet:
> I have a couple of questions left:
> 
> 1. Should we have a "padding"? If we use the min/max as lims then some points will be cut in half. Adding a small padding, 5-10% of the figure width prevents this but it might look weird in other cases, for example image plots doesnt benefit from padding. I think matplotlib does this depending on the plot type.

Padding on 1d plots: yes, Padding on 2d plots: no. (basically like MPL)

> 2. When the norm is 'log' and some y-value value is negative then the bounding box strategy we use fails. Preferably we would like the select the bottom limit as the minimum positive y-value, but when we have bounding boxes we can't access the minimum positive value. Right now I just pick a bottom value 0.1 x the top value. Should we do something else instead?

Can you use the `find_limits` to determine the bounding boxes? It is supposed to do exactly that: find the last positive value. However, you probably don't know if it's linear or log scale at that point? Unless you pass it as an argument to `bbox` when called from the canvas?

> 3. Should there be a toggle to turn on/off this new behavior? Should the new behavior be default or optional?

I would not add a toggle button for now.

I'm not sure what we want to do in terms of making users able to toggle this on or off. We probably need to be careful in adding too many options/arguments. Also, once you make a new option available, it is more difficult to remove it, because if people use it in their notebooks, it will break the notebooks.
So I would say just make this default for now, do not add a keyword arg to control it, and if people really complain then we can add it later?

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -174,6 +181,14 @@ def autoscale(self):
                 xmax = max(xmax, right.value)
                 ymin = min(ymin, bottom.value)
                 ymax = max(ymax, top.value)
+        if self.yscale == 'log' and ymin < 0:
+            # Something has gone wrong
+            # probably the y-limit was set before the scale changed
+            # set the y-limit to something reasonable
+            ymin = 1e-2 * ymax
+        if self.xscale == 'log' and xmin < 0:
+            xmin = 1e-2 * xmax
```
nvaytet, line 190: Hmm I don't like this so much. I think we should be able to come up with a robust way of doing things where this is not needed.

src/plopp/backends/matplotlib/line.py
```diff
@@ -229,6 +233,10 @@ def update(self, new_values: sc.DataArray):
                 )
             )
 
+    def _update_limits(self):
```
nvaytet, line 236: It is a bit unfortunate that we have to remember to update the limits every time the data changes. If we add later another function that modifies the data, we may forget to also call this.

src/plopp/core/limits.py
```diff
@@ -1,39 +1,64 @@
...
-    finite_inds = np.isfinite(v)
-    if np.sum(finite_inds) == 0:
+    m = np.ones_like(x.values, dtype='bool')
+    if isinstance(x, DataArray) and len(x.masks):
+        m &= ~merge_masks(x.masks).values
+    if not np.any(m):
+        m = np.ones_like(x.values, dtype='bool')
+
+    v = x.values[m]
+    if x.ndim == 1:
```
nvaytet, line 28: I don't think you can rely on the data being 1d, because the data sent to a `scatter3d` plot will also be 1d, and you will get a 1d data from the `colormapper`. In that case you don't want to take into account the variances.

src/plopp/core/limits.py
```diff
@@ -1,39 +1,64 @@
...
-            finite_min = np.amin(finite_vals, initial=initial, where=finite_vals > 0)
+            finite_min = np.amin(v, initial=initial, where=v > 0)
     else:
-        finite_min = np.amin(finite_vals)
+        finite_min = np.amin(v)
     if finite_max is None:
-        finite_max = np.amax(finite_vals)
+        finite_max = np.amax(v)
+    # Padding only for 1d plots
+    if x.ndim == 1:
```
nvaytet, line 53: For the same reason as above (data can be 1d for 3d plots), I think whether we apply padding or not should be decided by an argument to the function.

tests/core/limits_test.py
```diff
@@ -11,46 +11,46 @@
 def test_find_limits():
     x = sc.arange('x', 11.0, unit='m')
     lims = find_limits(x)
-    assert sc.identical(lims[0], sc.scalar(0.0, unit='m'))
-    assert sc.identical(lims[1], sc.scalar(10.0, unit='m'))
+    assert sc.identical(lims[0], sc.scalar(-0.5, unit='m'))
```
nvaytet, line 14: A little annoying that if we change the size of the padding, we have to change all the tests...

nvaytet (commented):
Have a look at #292 where I took a slightly different approach.
Does it solve most of the issues here?

We can decide which one to move forward with.

jokasimr:
Yes I think the PR you linked solves almost all of the issues you mentioned there, and the rest seem minor / easy to fix.

PR #292, opened 2024-01-10
Ignore masked values for determining canvas and colorbar ranges

### LINKED ISSUE #266: Allow other ways of handling masks?

nvaytet:
Fixes #266 

As always, it turned out more complicated than expected. I needed to handle a bunch of different cases with coordinates that contain strings or datetimes.

For the latter, logscale on a datetime axis currently does not crash but no padding is applied.
It is an edge case anyway, and renders funny in matplotlib. I don't think it is very useful as a plot.
At least it doesn't crash so users can anyway set their own limits.

### DIFF
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -10,7 +10,7 @@ from matplotlib.collections import QuadMesh
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
 from ...core.utils import maybe_variable_to_number, scalar_to_string
-from ..common import BoundingBox, get_canvas_bounding_box
+from ..common import BoundingBox, axis_bounds
 from .utils import fig_to_bytes, is_sphinx_build, make_figure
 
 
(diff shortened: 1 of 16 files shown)


jokasimr:
Yeah this looks like a better solution. The autoscaling logic doesn't leak into the `Line` class and we don't get the problem that the limits aren't updated when the scale is changed.

jokasimr:
About the padding breaking the tests, perhaps we should make the padding width a global config and then import it both in `limits.py` and in the tests.
This would also let the user tweak it if they want to.

src/plopp/core/limits.py
```diff
@@ -1,21 +1,60 @@
...
+    xmax: Optional[float] = None
+    ymin: Optional[float] = None
+    ymax: Optional[float] = None
+
+    def union(self, other: BoundingBox) -> BoundingBox:
+        """
+        Return the union of this bounding box with another one.
+        """
+        return BoundingBox(
+            xmin=min(self.xmin or np.inf, other.xmin or np.inf),
```
jokasimr, line 31: Wouldn't this break if `self.xmin == 0`? I think we need to explicitly check for `None` here even if it is more verbose. Or perhaps we can set the defaults already in the constructor?

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -112,11 +143,7 @@ def __init__(
         self.ax.set_title(title)
         self.ax.grid(grid)
         self._coord_formatters = []
-
-        self._xmin = np.inf
-        self._xmax = np.NINF
-        self._ymin = np.inf
-        self._ymax = np.NINF
+        self._bbox = BoundingBox()
```
jokasimr, line 146: Would it be possible to have this only be a local variable in the `autoscale()` method? I saw that it is used in `logy`  but I didn't really understand why.

jokasimr (commented):
Looks good to me! Just have some minor comments.

src/plopp/backends/matplotlib/canvas.py
nvaytet: It's also used in `autoscale` itself. It's used as a global memory of the range. It is needed when you are in a mode where you allow the plot limits to grow but not shrink.
The default is that the range follow the data, but you can also set a mode where limits grow, maybe giving a better feel for the relative differences in the data.

See the `autoscale` parameter in https://scipp.github.io/plopp/reference/generated/plopp.plot.html#plopp.plot

nvaytet:
I now realise I also need to apply this to the plotly backend.

src/plopp/core/limits.py
```diff
@@ -1,51 +1,86 @@
...
+    scale: Optional[Literal['linear', 'log']] = 'linear',
+    pad: bool = False,
+) -> Tuple[sc.Variable, sc.Variable]:
     """
     Find sensible limits, depending on linear or log scale.
     If there are no finite values in the array, raise an error.
     If there are no positive values in the array, and the scale is log, fall back to
     some sensible default values.
     """
+    if scale is None:
```
jokasimr, line 23: Minor: would it be nicer to leave `scale` as `linear/log` here and let the caller worry about the `None` case?

src/plopp/core/limits.py
```diff
@@ -1,51 +1,86 @@
...
+from typing import Literal, Optional, Tuple
 
 import numpy as np
-from scipp import Variable, scalar
+import scipp as sc
+
+from .utils import merge_masks
 
 
 def find_limits(
```
jokasimr, line 12: Is `find_limits` called anywhere with a `Variable` valued argument for `x`?

jokasimr (approved):

PR #293, opened 2024-01-16
Fix plotting with errorbars when nan values are present

nvaytet:
This used to fail after #292 

```Py
import plopp as pp
import numpy as np

da = pp.data.data1d(variances=True)
da.values[-10:] = np.nan
da.plot()
```

### DIFF
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -169,11 +169,17 @@ class Canvas:
                 bbox = bbox.union(mesh_bbox)
 
             elif hasattr(c, '_plopp_mask'):
-                line_mask = sc.array(dims=['x'], values=c._plopp_mask)
+                segments = c.get_segments()
+                # Here get_segments() can return empty segments in the case where the
+                # data values are NaN, which we filter out.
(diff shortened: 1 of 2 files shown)


src/plopp/backends/matplotlib/canvas.py
```diff
@@ -169,11 +169,17 @@ def autoscale(self):
...
+                segments = c.get_segments()
+                # Here get_segments() can return empty segments in the case where the
+                # data values are NaN, which we filter out.
+                lengths = np.array([len(segs) for segs in segments])
+                line_mask = sc.array(dims=['x'], values=c._plopp_mask[lengths > 0])
                 line_y = sc.DataArray(
                     data=sc.array(
                         dims=['x', 'y'],
-                        values=np.array(c.get_segments())[..., 1],
+                        values=np.array([s for (s, l) in zip(segments, lengths) if l])[
```
jokasimr, line 180: I think this would be more clear by doing something like:
```python
np.array(c.get_segments())[..., 1][lengths > 0]
```
if that is possible. It would make it more obvious that we are using the same criteria to filter the mask and the values.

PR #296, opened 2024-01-19
Add polar plot

nvaytet:
Fixes #288

PR #298, opened 2024-01-24
Better handling of polar axes

### LINKED ISSUE #288: Add polar plot functionality
Currently, I have to convert everything into numpy arrays and use basic Matplotlib methods to generate polar plots, which strips away the awesome features in Plopp/Scipp.  How easy/hard would this be to add (including unit detection, error bars, etc)?  Maybe just a kwarg like is done in plt.subplots(subplot_kw={'projection':'polar'})?  It isn't too burdensome right now, but makes my code more messy/less readable for students.

nvaytet:
Looking at how many changes were required in #296 to support polar plots, I am very worried that it is a bad case of feature creep.

Instead, the same plots can be achieved with a very small mix of matplotlib and plopp code.
I suggest to simply add a notebook in the gallery to show how to make polar plots, which are helped by a couple of tweaks here in our `Canvas` class.

Fixes #288


jl-wynen (approved):
Looks good!

I guess for convenience, we could have a `pp.polar` function that creates an axes as shown in the example. This doesn't seem like it would require changes to existing code. But this can be added later if it's needed.

nvaytet:
> I guess for convenience, we could have a pp.polar function that creates an axes as shown in the example. This doesn't seem like it would require changes to existing code. But this can be added later if it's needed.

Yes, the issue with doing that is that you would be inserting Matplotlib specific in the top level functions of scipp.
It wouldn't work with a different backend (e.g. `plotly`), which was the reason why you needed all the code in the other PR.

But I don't really feel like adding all those changes in, for a result that still doesn't look great.
For instance, there is still a unit of radians `[rad]` placed on one of the axes, which doesn't make much sense because the numbers on the ticks are in degrees.

The author of the issue says that they are losing a lot of the plotting functionality by having to go through numpy and raw mpl, but I am not so sure how much is really lost. Even interactive plots should still be doable using this method.

TheFermiSea:
Hello @nvaytet,
I am thankful for your modifications and examples here.  For basic plotting, this works great.  However, I am now attempting to build a custom interface similar to the example for Rectangle Selection in the user guide.  It seems that whenever the canvas is updated (for instance, the RectangleTool selects a slicing range), the polar axis limits are reset to some strange values instead of [0,2*pi].   This is also a problem with basic plotting, but the workaround there is to simply call ax.set_thetalim([0., 2*np.pi]) before calling the created figure (as shown in the Polar Plot example).  Is there a way to insert an update node into the graph that forces the axis limits?

nvaytet:
Hi, I have noticed the resetting of axes upon updates, and I think it is the same issue as in #340 .
I will work on a fix.

PR #300, opened 2024-01-25
Copier update

nvaytet:


jl-wynen:
Sounds like `custom.in` should be a constraints file (https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-c).

This approach makes the setup even more complex. Did you add the extra files specifically so you don't have to duplicate the optional dependencies in `base.in` and `nightly.in`?

nvaytet:
> Did you add the extra files specifically so you don't have to duplicate the optional dependencies in base.in and nightly.in?

Yes.

nvaytet:
> Sounds like custom.in should be a constraints file (https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-c).

Never used one. Does that mean that everywhere I use `pip` I need to add the `-c` flag?

jl-wynen:
> Never used one. Does that mean that everywhere I use pip I need to add the -c flag?

Me neither. But it should mean that instead of `-r custom.in` it should be `-c custom.in` in the requirements files.

nvaytet:
See update

jl-wynen (approved):

PR #304, opened 2024-02-02
Remove run dependency on scipp in conda

### LINKED ISSUE #302: Conda package depends on scipp
We did not want the packages to depends on scipp so that we can use plopp in dev builds of scipp. But since plopp 23.11, the conda package does seem to depend on scipp. I don't know where this comes from, though, because scipp is not listed as a run dependency in `meta.yaml`. Maybe something happens with the test dependency.

jl-wynen:
Fixes #302

Do we still need a `run_constrained`? There is none in pyproject.toml...

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -16,7 +16,6 @@ requirements:
   run:
     - matplotlib
     - python>=3.9
-    - scipp
 
 test:
   imports:


nvaytet:
> Do we still need a run_constrained? There is none in pyproject.toml...

I don't know really. I guess it's either we add one in both places or nowhere?

jl-wynen:
> > Do we still need a run_constrained? There is none in pyproject.toml...
> 
> I don't know really. I guess it's either we add one in both places or nowhere?

Do you know the minimum version? if not, I'd just leave it as is. It's rare that users get an old version anyway.

nvaytet (approved):

PR #305, opened 2024-02-05
Update dependencies to fix nightly build

nvaytet:


jl-wynen (approved):
Which one is the problem?

nvaytet:
> Which one is the problem?

Sorry, I updated the PR description.

jl-wynen:
That should have been fixed in the template. Did you update?

nvaytet:
> That should have been fixed in the template. Did you update?

I think it's fine in the template. I have a feeling it is the result of some manual editing, probably in #300

jl-wynen (approved):

PR #307, opened 2024-02-14
Plot intensity on regular polyhedra

g5t:
Position-sensitive <sup>3</sup>He detectors are (typically) cylinders with cylindrical-axis-length much greater than their tube diameter. Through charge division the position of a detection event can be determined continuously, but due to finite resolution of the underlying physics and analog-to-digital conversion systems the reconstruction position is typically pixelated.

This type of detector is common enough to have a special representation of its geometry as part of the NeXus specification, the [NXcylindrial_geometry](https://manual.nexusformat.org/classes/base_classes/NXcylindrical_geometry.html#nxcylindrical-geometry) -- in which each cylindrical pixel is represented by three vertices.

This PR adds `plopp.cylinders3d` to plot a `scipp.DataGroup` containing the information from an `NXdetector` with `NXcylindrical_geometry` pixel information.

## Implementation
The `DataGroup` to be displayed must contain 
- the intensity `data` as a 1-D `scipp.Variable` with `N` entries
- each cylinder's three-vertex indices must be stored in a `N` by `3` 2-D `scipp.Variable`, `cylinders`; to be valid indexes the stored integers should all be positive
- the `vertices` should be a 1-D `scipp.Variable` with `scipp.DType('vector3')` data type and _at least_ `cylinders.max()+1` total vertices

The information must be contained in a `DataGroup` because there is no strict relationship between the number of cylinders and the number of vertices -- vertices can be shared between cylinders, and extra vertices are allowed.

Compared to displaying the same intensity data as a `plopp.scatter3d` at each cylinder's center of mass,
using `plopp.cylinders3d` allows for an exact representation of the size and shape of the cylindrical pixels 
at the expense of longer pre-display computations.

## Possible alternate uses
If the number of `radial_segments` is set to `2` and `open_ended` is set to `True`, each cylinder is rendered as a two-sided rectangle (four vertices and four triangulated faces) defined by the three 'cylinder' vertices.
This can produce faster rendering times _and_ planar  representations of pixel sizes for cylinders or any rectangular-shaped cross-sectioned detector pixels.

## Possible issues
I've tried to fit this new functionality into the existing `plopp` structure, but may have done so poorly.
The requirement that input-data be contained in a `scipp.DataGroup` seems to be at-odds with other plotting routines, which split `DataGroup` entries into separate plots.


docs/gallery/teapots.ipynb
```diff
@@ -0,0 +1,105 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "17405423-0496-40d5-973c-c7ec6aa42746",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def make_one_teapot_mesh3d(**kwargs):\n",
+    "    mesh, faces, names = teapot()\n",
+    "    return pp.mesh3d(mesh, faces, **names, **kwargs)"
```
nvaytet, line 25: So this is different from the interface I had in mind.
I know it is modeled on the interface to `scatter3d`, but I am starting to wonder if we should also change that.

I thought one could maybe have something that would resemble more what the mesh needs, like
```Py
pp.mesh3d(vertices=sc.array(shape=N, dtype=vector3),
          faces=sc.array(shape=[N, 3]),
          color=sc.array(shape=N, dtype=float))
```
Internally, a bit like [xyplot](https://github.com/scipp/plopp/blob/main/src/plopp/plotting/xyplot.py), the `vertices, faces, color` inputs could be wrapped into nodes, before supplying them to the figure, so that if one of them updates, it would notify the figure.

Only downside is that this makes it more difficult to plot multiple meshes (on scatter3d, you can supply a dict of data arrays to plot two point clouds on the same scene).

Not sure what a good API would be for that, but I would like to use a similar API for the normal scatter plot:
```Py
pp.scatter3d(x=array_x, y=array_x, z=array_x, color=array_data)
```

nvaytet: Is there a reason why we the mesh needs to be made of N identical objects? Sure that's what it would be like for detector pixels of an instrument, but really if you are making something that can plot a mesh, you should be free from that restriction?

I was seeing it as something just very generic, that would show a mesh. The fact that we construct it with `NxV` vertices would be done externally (e.g. in `essspectroscopy`).
Then the colors for the vertices would also need to be constructed there (also `NxV` in size).

That said, this would then make it much more difficult to get it to work with the plane slicing.
I guess for the bifrost instrument view, having such slicing would be useful?
One could say that if you really need the slicing, you have to stick to the simple point cloud.

g5t: My initial thought seeing this was that `N=1` is a special case already handled by the implementation that could be extended to an _implicit_ `N=1` if that dimension is missing from the input.
But I think I now understand your point, and may even understand the motivation for allowing the three vector components of the vertices to be specified separately.

---

My concern is that plotting intensity on `N` topologically-equivalent polyhedra is not a BIFROST, spectroscopy, or cylindrical-detector-instrument specific task; and that converting from `N` &times; `V` vertices and `N` intensities to their flattened lists is not _entirely trivial_, especially for someone unfamiliar with the dimension traversal order of `scipp.flatten`.

Would the `plopp`-way of achieving this be an extra node to handle the conversion before your desired-signature `mesh3d`?
Can that node be part of `plopp` to avoid duplication and the chance for errors that brings?

nvaytet:
See #346

nvaytet:
Closing as meshes were introduced in #376 instead.
Feel free to open a new issue if you need support for N identical meshes.

PR #311, opened 2024-02-22
Fix cursor hover when datetime axis is present in figure

### LINKED ISSUE #310: Cursor hover is broken on image with datetime coord
To reproduce:
```Py
import plopp as pp
import numpy as np
import scipp as sc
%matplotlib widget

t = np.arange(
    np.datetime64('2017-03-16T20:58:17'), np.datetime64('2017-03-16T21:15:17'), 20
)
time = sc.array(dims=['time'], values=t)
z = sc.arange('z', 50.0, unit='m')
v = np.random.random(z.shape + time.shape)
da = sc.DataArray(
    data=sc.array(dims=['z', 'time'], values=10 * v, variances=v),
    coords={'time': time, 'z': z},
)
pp.plot(da)
```

Then, hovering the mouse cursor over the image raises:
```
File ~/code/plopp/jupyter/plopp/backends/matplotlib/image.py:228, in Image.format_coord(self, x, y)
    220 try:
    221     val = self._data_with_bin_edges[
    222         self._hover_slicing['dim'][0],
    223         sc.scalar(
   (...)
    226         ),
    227     ]
--> 228     val = val[
    229         self._hover_slicing['dim'][1],
    230         sc.scalar(
    231             xy[self._hover_slicing['dir'][1]],
    232             unit=self._hover_slicing['unit'][1],
    233         ),
    234     ]
    235     prefix = self._data.name
    236     if prefix:

DTypeError: 'less_equal' does not support dtypes 'datetime64', 'float64', 
```
(set log level to Debug in log console to see the error).

It also means that when trying to zoom, no zoom rectangle is drawn.

nvaytet:
Fixes #310

### DIFF
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -29,6 +29,9 @@ def _cursor_value_to_variable(
     x: Union[float, int], dtype: sc.DType, unit: str
 ) -> sc.Variable:
     if dtype == sc.DType.datetime64:
+        # Annoying chain of conversion but matplotlib has its own way of converting
+        # dates to numbers (number of days since epoch), and num2date returns a python
+        # datetime object, which scipp expects a numpy datetime64.
         return sc.scalar(np.datetime64(mdates.num2date(x).replace(tzinfo=None))).to(
             unit=unit
         )
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -15,7 +15,7 @@ from ..common import BoundingBox, axis_bounds
 from .utils import fig_to_bytes, is_sphinx_build, make_figure
 
 
-def _to_floats(x):
+def _to_floats(x: np.ndarray) -> np.ndarray:
     return mdates.date2num(x) if np.issubdtype(x.dtype, np.datetime64) else x
 
 
@@ -23,6 +23,12 @@ def _none_if_not_finite(x: Union[float, int, None]) -> Union[float, int, None]:
     if x is None:
         return None
     return x if np.isfinite(x) else None
+
+
+def _cursor_formatter(x: Union[float, int], dtype: sc.DType, unit: str) -> str:
+    if dtype == sc.DType.datetime64:
+        return mdates.num2date(x).replace(tzinfo=None).isoformat()
+    return scalar_to_string(sc.scalar(x, unit=unit))
 
 
 class Canvas:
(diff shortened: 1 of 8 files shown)


nvaytet:
@jl-wynen would be super if you could have a quick look at this

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -25,6 +25,25 @@ def _none_if_not_finite(x: Union[float, int, None]) -> Union[float, int, None]:
...
 
 
+def _cursor_value_to_variable(
+    x: Union[float, int], dtype: sc.DType, unit: str
+) -> sc.Variable:
+    if dtype == sc.DType.datetime64:
+        # Annoying chain of conversion but matplotlib has its own way of converting
+        # dates to numbers (number of days since epoch), and num2date returns a python
+        # datetime object, while scipp expects a numpy datetime64.
+        return sc.scalar(np.datetime64(mdates.num2date(x).replace(tzinfo=None))).to(
```
jl-wynen, line 35: We typically say that scipp datetimes are in UTC. But this looks like the result will be in whatever timezone matplotlib uses but without timezone info. I am assuming mpl uses the local timezone unless specified otherwise in the input. Is that correct?

Did you check that the cursor label is consistent with the axis labels regardless of timezone? And the same for zooming, selecting rectangles, etc.

src/plopp/backends/matplotlib/image.py
```diff
@@ -132,29 +133,6 @@ def __init__(
         if need_grid:
             self._ax.grid(True)
 
-        # Cache slicing order for hover values
```
jl-wynen, line 135: Is this related to datetimes?

nvaytet: Kind of, it was the old way of getting the values for the fomatter. This is where the units for each axis were used from, but now, we instead just use the units that are stored in the canvas. The info was basically stored in both places and I think it makes more sense to just have it on the canvas.

We basically need to slice the image data to get the z value. We use label-based slicing, and for that we need the dim associated to each axis, and the unit. The canvas knows both these things, and the canvas now just sends a tuple `(dim, sc.scalar(value, unit=unit)` to the `format_coord` which is used as is for slicing.

src/plopp/backends/matplotlib/canvas.py
nvaytet: I think that matplotlib defaults to UTC:

> All the Matplotlib date converters, locators and formatters are timezone aware. If no explicit timezone is provided, `rcParams["timezone"]` (default: 'UTC') is assumed,

in https://matplotlib.org/stable/api/dates_api.html#matplotlib-date-format.

In the examples I used, the first point read from the cursor corresponds to the first value in the array:
Screenshot at 2024-03-08 13-39-25

jl-wynen: Ok. And zooming and selecting points with mpl toolbox all work?

nvaytet: Seems like it, yes
Figure 1 (3)

jl-wynen (approved):

PR #312, opened 2024-02-22
Better datetime axis compatibility

nvaytet:
Computing limits on datetime axes is a little funny with data has that passed through matplotlib.
Datetimes are converted to floats for pcolormesh coordinates (2d images) but not for lines.

This PR tries to handle the different cases in a more consistent way, which allows us to plot 1d and 2d data on the same axes with datetime coords (which failed previously). It is not a common use case but was requested by users.

Example:
```Py
import plopp as pp
import numpy as np
import scipp as sc

# 2d data
t = np.arange(
    np.datetime64('2017-03-16T20:58:17'), np.datetime64('2017-03-16T21:15:17'), 20
)
time = sc.array(dims=['time'], values=t)
z = sc.arange('z', 50.0, unit='m')
v = np.random.random(z.shape + time.shape)
da = sc.DataArray(
    data=sc.array(dims=['z', 'time'], values=10 * v, variances=v),
    coords={'time': time, 'z': z},
)
f = pp.plot(da)

# 1d data
v = np.random.rand(time.sizes['time'])
a = sc.DataArray(
    data=sc.array(dims=['time'], values=10 * v, variances=v),
    coords={'time': time},
)
a.plot(ax=f.ax, color='r')
```
Screenshot at 2024-02-22 16-51-54

### DIFF
--- a/src/plopp/backends/common.py
+++ b/src/plopp/backends/common.py
@@ -6,9 +6,9 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Dict, Literal, Optional, Tuple
 
-from matplotlib import dates as mdates
 import numpy as np
 import scipp as sc
+from matplotlib import dates as mdates
 
 from ..core.limits import find_limits, fix_empty_range
 from ..core.utils import merge_masks
--- a/src/plopp/backends/common.py
+++ b/src/plopp/backends/common.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Dict, Literal, Optional, Tuple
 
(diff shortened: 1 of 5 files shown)


src/plopp/backends/common.py
```diff
@@ -8,17 +8,22 @@
 
 import numpy as np
 import scipp as sc
+from matplotlib import dates as mdates
```
jl-wynen, line 11: This is in the `common` backend module. Shouldn't it be in the matplotlib module?

nvaytet: Hmm, good point

jl-wynen (approved):

PR #314, opened 2024-02-26
Replace cut slices with clipping planes in scatter3d

### LINKED ISSUE #313: Bad instrument view button formatting
In the rendered docs, the 'toggle outline visibility' button of instrument views has a bad icon:
Screenshot_20240125_111730
It alternates between an exclamation mark and question mark.

nvaytet:
The old method of making 3d cuts was slightly strange in the way that selection of points to display was based around a plane that had a thickness. There were buttons for increasing/decreasing the thickness, but the user could not really see the actual thickness so well.

In this PR, we have re-worked the tool to use two planes instead of one, to define a range for selection. The slider is not a range-slider with two handles. Note that both handles can be dragged at the same time by clicking in between the two handles.
Screenshot at 2024-02-26 09-59-49

**Notable changes:**

- The sliders are now arranged in tabs
- Allows multiple cuts in the same (X, Y, or Z) direction
- Has a drop-down menu to define which operation to use to combine the slices (OR by default, as in the old tool)

I kept the old tool, for backwards compatibility. I don't know if anyone was using it on its own.
I don't know if we should add a deprecation warning saying it will no longer be maintained?

Also fixes #313


nvaytet:
We should probably get #317 merged first, to avoid too many merge conflicts the other way round.

src/plopp/widgets/clip3d.py
```diff
@@ -0,0 +1,395 @@
...
+        for outline in self.outlines:
+            outline.visible = value
+
+    def move(self, value: Dict[str, Any]):
+        """
+        Move the outline of the cut according to new position given by the slider.
+        """
+        # Early return if relative difference between new and old value is small
+        old = np.array(value['old'])
+        if np.abs((np.array(value['new']) - old) / old).max() < 1e-6:
```
jokasimr, line 163: What are `value['new']` and `value['old']` here?

jokasimr: I'm thinking that if they are slider values and the interval is something like `[1000_000, 1000_001]` then this condition might always be true. But if they're always small then no problem. Why is the check needed? We're debouncing / throttling anyway right?

jokasimr (commented):
Havn't finished the review yet. Will continue tomorrow.

docs/api-reference/index.md
```diff
@@ -81,6 +81,8 @@
    widgets.drawing.DrawingTool
    widgets.drawing.PointsTool
 
+   widgets.clip3d.Clip3dTool
+   widgets.clip3d.ClippingPlanes
    widgets.cut3d.Cut3dTool
```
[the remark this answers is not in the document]
nvaytet: Yes, it's the old tool.
I was thinking of keeping it in case someone was using it.
But you're right, I should probably deprecate it, or at least indicate that we will not be maintaining it.

The alternative is of course just to remove it...

nvaytet:
> Havn't finished the review yet. Will continue tomorrow.

@jokasimr did you have more comments?

src/plopp/widgets/clip3d.py
nvaytet: This was serving a double purpose where it would also prevent flickering of an existing cut when adding a new one. I updated the comment and changed the threshold so that it is based on the magnitude of the step of the slider (which itself is based on the range of the slider).

jokasimr:
> @jokasimr did you have more comments?

Sorry I haven't looked at the rest yet. I'll try to finish it up later this afternoon.

docs/api-reference/index.md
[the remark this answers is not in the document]
jokasimr: Either works for me, I think it's best you decide

src/plopp/widgets/clip3d.py
```diff
@@ -0,0 +1,397 @@
...
+
+    Parameters
+    ----------
+    limits:
+        The spatial extent of the points in the 3d figure in the XYZ directions.
+    direction:
+        The direction normal to the slice.
+    plain_update:
+        A function to update the scene without debouncing.
+    throttled_update:
```
jokasimr, line 53: Doesn't it always makes sense to have some debouncing? For example, is there ever any use running the function more frequently than the screen refreshes?

In that case, maybe it's a nice api to just have one argument for the update function, and then a `debounce_time` argument with default `0.1` or something like that?

src/plopp/widgets/clip3d.py
```diff
@@ -0,0 +1,397 @@
...
+
+    .. versionadded:: 24.04.0
+
+    Parameters
+    ----------
+    limits:
+        The spatial extent of the points in the 3d figure in the XYZ directions.
+    direction:
+        The direction normal to the slice.
+    plain_update:
```
jokasimr, line 51: What's the function signature of the update function? Unless it's something that's already written elsewhere that should probably be part of the docs here.

src/plopp/widgets/clip3d.py
```diff
@@ -0,0 +1,397 @@
...
+            layout={'width': '16px', 'padding': '0px'},
+        )
+
+        for outline, val in zip(self.outlines, self.slider.value):
+            pos = list(center)
+            pos[axis] = val
+            outline.position = pos
+
+        self.unit_label = ipw.Label(f'[{self._unit}]')
+        self.cut_visible.on_click(self.toggle)
```
jokasimr, line 132: Bug we saw in huddle: When `cut_visible` is toggled the outline is shown even if all outlines are hidden.

jokasimr (approved):

PR #317, opened 2024-03-04
Two-dimensional scatter plot

### LINKED ISSUE #264: Add a scatter (2d) plot?
Would look like:

img

### LINKED ISSUE #315: Make use of figure side bars instead of returning Box
In several high-level plots (`scatter3d`, `superplor`, `slicer`) we are returning a `Box` that contains, e.g. the `Figure` and a `slider` below (in the case of the `slicer` plot).
Instead, we should make use of the figure's `bottom_bar`, which is a container for such widgets, so that the returned object is a figure, thus more consistent with the other plotting functions.

### LINKED ISSUE #323: Display on super-plot somewhat broken after #319
After #319 we now see this:
Screenshot at 2024-04-08 15-09-02

The figure is cropped.

I think it's something specific to the matplotlib canvas, which is messing up when we add more widgets to the right.
It seems it taking 80% of the box it's currently in, not 80% of the page :-(

nvaytet:
This PR introduces a new 2d scatter plot.
Fixes #264 

Example:
```Py
import plopp as pp

a = pp.data.scatter()
pp.scatter(a, cbar=True)
```

See docs notebook for more examples.

In addition, there has been a refactor of the interface to how we update figures:
Old:
```Py
fig.update(data_a, key='a')
fig.update(data_b, key='b')
```
New:
```Py
fig.update(a=data_a, b=data_b)
```
It made sense to do this change here to unify the different views (made a common `GraphicalView` base class).

Also fixes #315
Also fixes #323


src/plopp/backends/common.py
```diff
@@ -91,27 +91,24 @@ def make_line_data(data: sc.DataArray, dim: str) -> dict:
...
-        if mask is not None:
-            mask['y'] = np.concatenate([mask['y'][0:1], mask['y']])
-    return {
-        'values': {'x': x.values, 'y': y.values},
-        'stddevs': error,
-        'mask': mask,
-        'hist': hist,
-    }
+        for array in (values, mask):
+            array['y'] = np.concatenate([array['y'][0:1], array['y']])
```
jokasimr, line 113: What changed here? Before we didn't add a point to the values, but now we do.

src/plopp/backends/matplotlib/interactive.py
```diff
@@ -26,7 +26,13 @@ def __init__(self, View, *args, **kwargs):
         super().__init__(
             [
                 self.top_bar,
-                HBox([self.left_bar, self._view.canvas.to_widget(), self.right_bar]),
+                HBox(
+                    [
+                        self.left_bar,
+                        VBox([self._view.canvas.to_widget()]),
```
jokasimr, line 32: I don't understand what this does, why is the widget wrapped in a `VBox`?

src/plopp/graphics/colormapper.py
```diff
@@ -227,9 +227,10 @@ def _set_normalizer_limits(self):
             self.normalizer.vmin = self.vmin
             self.normalizer.vmax = self.vmax
 
-    def update(self, key: str, data: sc.DataArray):
+    def update(self, _=None, **kwargs):
```
jokasimr, line 230: The argument name confused me here, I always assume `_` is unused, so I was surprised when it was used.
Maybe we should rename it to `values`?

src/plopp/graphics/graphicalview.py
```diff
@@ -0,0 +1,84 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
+
+from ..core import Node, View
+from ..core.utils import make_compatible, name_with_unit
+
+
+class GraphicalView(View):
+    """ """
```
jokasimr, line 9: I think this needs a docstring explaining what makes it different from `View`.

src/plopp/graphics/lineview.py
```diff
@@ -102,64 +103,25 @@ def __init__(
...
-
-        Parameters
-        ----------
-        new_values:
-            New data to create or update a :class:`Line` object from.
-        key:
-            The id of the node that sent the new data.
-        """
-        if new_values.ndim != 1:
-            raise ValueError("LineView can only be used to plot 1-D data.")
```
jokasimr, line 128: Maybe I missed something but it seems we loose this error message here. Is there another check elsewhere using the new `_ndim` attribute?

jokasimr (commented):
Looks good to me. Tested the scatter plot and played around with it a bit and it seems to work well. The changes to the `update` method and the to the structure seem to make a lot of sense.

The superplot was not interactive for me, do I need another package to make it work?

src/plopp/backends/common.py
nvaytet: I think what changed is that before we were using `scipp.concat` for y values, and `numpy` for the mask values, but now we switch to numpy earlier (because I had to resort to using `np.asarray` because in some cases some lists of arrays were supplied)

src/plopp/graphics/colormapper.py
nvaytet: I basically wanted to avoid a name clash from happening. Like when you update a dict in python.
You can do
```Py
d.update(a=1, b=3, values=5)
d.update({'a': 1, 'b': 3, 'values': 5})
```
If my function signature is
```Py
def update(self, values=None, **kwargs):
```
then the first case will fail because it is expecting `values` to be a dict, not just a single value.

I don't actually know how python does it, because you can also use `_` as a key:
```Py
d = {'a': 1, 'b': 3}
d.update(_=55)
print(d)
d.update({'a': 33, 'c': 6})
print(d)
```
```
{'a': 1, 'b': 3, '_': 55}
{'a': 33, 'b': 3, '_': 55, 'c': 6}
```
Maybe I should look at how they implemented it...

nvaytet: So I think a better way of implementing this is
```Py
def update(self, *args, **kwargs):
    new = kwargs
    if args:
        if len(args) > 1:
            raise TypeError('Update expected at most 1 arg.  Got {}.'.format(len(args)))
        else:
            new.update(args[0])
    ...
```

jokasimr: No I think what you did makes sense. We can avoid clashes and make the name more clear by using something like: `_dict_arg`. That's unlikely to cause a clash and it's also a bit more clear. But I'll leave it to you to decide.

jokasimr:
> > The superplot was not interactive for me, do I need another package to make it work?
> 
> Did you use the notebook from the docs? Did you have `%matplotlib widget` enabled?

Yes I did. But when running it the first time it told me to install `ipympl`, so I had to do that manually before running, then it ran but was not interactive. Maybe that's related to the issue.

src/plopp/graphics/lineview.py
nvaytet: Yes, at the start of the `update` function in the `GraphicalView` ;-)

src/plopp/backends/matplotlib/interactive.py
nvaytet: This was an annoying thing where the buttons would end up all the way to the right, and for some reason, wrapping it in a `VBox` helps.
However, I now see that's it's still a little broken (see #323).

nvaytet:
I think you need to skip the weird cell with
```Py
from ipywidgets import HBox

b = p.children[0].children[1].children[0]
sl = p.children[1].controls['y']['slider']
b.click()
sl.value = 20
b.click()
sl.value = 30
f = p.children[0].children[0]
f.children = [
    f.top_bar,
    HBox([f.left_bar, f.canvas.to_image(), f.right_bar]),
    f.bottom_bar,
]
```
which is hidden in the docs.

jokasimr:
> I think you need to skip the weird cell with

Yes that did the trick. How do you make it hidden in the docs?

nvaytet:
> Yes that did the trick. How do you make a cell hidden in the docs?

https://nbsphinx.readthedocs.io/en/0.9.3/hidden-cells.html

src/plopp/backends/matplotlib/interactive.py
nvaytet: I think i fixed it.

src/plopp/graphics/colormapper.py
nvaytet: Even with `*args` there is no clash, you can still do `view.update(args=1, a=2)` and the keys will be `args` and `a`.

Issue #169, opened 2023-02-27
Toolbar disappears when 3D canvas is too big


nvaytet:
Using a figsize which is larger than your window width causes the toolbar with buttons to disappear. This also happens if you hit the increase size button too many times.
This is a behavior from the widgets: the toolbar is actually still there, but it's just hidden behind the canvas.
Note that the color scale also disappears.

PR #319, opened 2024-04-04
FIX: Add a hard coded width bound to the p3js canvas

MridulS:
Fixes https://github.com/scipp/plopp/issues/169 (a hack-y one).

After some researching around about bouding canvases out of py3js it seems like this is the easiest
way of fixing the overflow problem of the canvas. A better solution would be to expose max_width from
py3js but I guess this is a minimal invasive change.

This does hard code the 80% width value but I guess we can expose that upstream to be user controllable
if the need arises.

### DIFF
--- a/src/plopp/backends/pythreejs/canvas.py
+++ b/src/plopp/backends/pythreejs/canvas.py
@@ -64,6 +64,7 @@ class Canvas:
         )
 
     def to_widget(self):
+        self.renderer.layout = ipw.Layout(max_width='80%', overflow='auto')
         return self.renderer
 
     def set_axes(self, dims, units, dtypes):


MridulS:
The failure seems unrelated. I can't restart the runner.

nvaytet:
> The failure seems unrelated. I can't restart the runner.

I now gave you the correct permissions, and restarted the build as well ;-)

src/plopp/backends/pythreejs/canvas.py
```diff
@@ -64,6 +64,7 @@ def __init__(
         )
 
     def to_widget(self):
+        self.renderer.layout = ipw.Layout(max_width='80%', overflow='auto')
```
nvaytet, line 67: The fix actually work very nicely, and as you say is non-invasive.

Any reason why you set the layout every time `to_widget` is called (which is when the repr for the canvas is requested) as opposed to passing the layout once to the `Renderer` constructor above?

MridulS:
> Edit: it seems the failure is consistent. I'm wondering if our CI setup is broken for PRs from forks?? I'll investigate.

Yup, it is indeed. Seems like an issue people run in https://github.com/actions/checkout/issues/551

src/plopp/backends/pythreejs/canvas.py
MridulS: I just want to localize this as much as possible as I don't really like this hard coded behaviour. We only deal with this problem (right now) when someone tries to bound the canvas generated next to some other widget layout. The `to_widget` call seems like an explicit point where the underlying canvas will interact with other widgets.

nvaytet (approved):

nvaytet:
I wonder if this would also remove the need for having e.g. `figsize=(550, 400)` in the docs, which (I think?) were there to prevent the figure from being too wide so as to not partially/totally hide the toolbar buttons? See for example [here](https://scipp.github.io/plopp/user-guide/plot-types/scatter3d-plot.html).

nvaytet:
I think the same fix could also apply to 2d figures, when using the `%matplotlib widget` backend, in the (rare) case where you would make a very wide figure. I think it would also hide the toolbar buttons without it? Adding it [here](https://github.com/scipp/plopp/blob/main/src/plopp/backends/matplotlib/canvas.py#L161) should work?

MridulS:
@nvaytet updated for 2d canvas and removed the use of figsize in 3d examples.

nvaytet (approved):
Needs #320 for builds to pass

PR #320, opened 2024-04-04
FIX: Build docs with branches of forks too

MridulS:
Let's see if this fixes https://github.com/scipp/copier_template/issues/151

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -50,6 +50,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ inputs.branch == '' && github.ref_name || inputs.branch }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
           fetch-depth: 0  # history required so cmake can determine version
       - uses: actions/setup-python@v4
         with:


MridulS:
Yeah this fixes building from forks

.github/workflows/docs.yml
```diff
@@ -50,6 +50,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ inputs.branch == '' && github.ref_name || inputs.branch }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
nvaytet, line 53:
```suggestion
          # see https://github.com/actions/checkout/issues/551#issuecomment-1113089390
          repository: ${{ github.event.pull_request.head.repo.full_name }}
```

MridulS: Shouldn't this be synced up with copier template too then?

nvaytet (approved):

.github/workflows/docs.yml
nvaytet: This was a suggestion, you can of course take it or leave it.

Issue #277, opened 2023-11-14
Add a range slicer widget


nvaytet:
Currently we have a `SlicerWidget` which is composed of `ipywidgets.IntSlider`s, typically to slice dimensions of data arrays.
We should also add a `RangeSlicerWidget` which would be very similar, but the slider would be a [`IntRangeSlider`](https://ipywidgets.readthedocs.io/en/8.1.2/examples/Widget%20List.html#intrangeslider) instead.

Note that we probably should also add a helper function for range slicing dims to go with it, i.e. the analog to [slice_dims](https://github.com/scipp/plopp/blob/main/src/plopp/widgets/slice.py#L100).

PR #322, opened 2024-04-05
ENH: Widget to slice over a range

MridulS:
This is one take on https://github.com/scipp/plopp/issues/277, this adds a new widget `RangeSliceWidget` which uses `IntRangeSlider` to select a slice of range.

The challening bit is how to select the operation on how to slice and show the data, I have added a function `range_slice_dims_sum` which sums along the dimension. This can easily be some other operation (count, average, flatten(?) ), I have implemented a sum operation for now as it made sense in the context of the example in the gallery (`scatter3d-with-slider.ipynb`).

Open to suggestions here, we should probably make the range slicing operation use configurable.


src/plopp/widgets/slice.py
```diff
@@ -112,3 +140,23 @@ def slice_dims(data_array: sc.DataArray, slices: Dict[str, slice]) -> sc.DataArr
...
+    Parameters
+    ----------
+    data_array:
+        The input data array to be sliced.
+    slices:
+        Dict of slices to apply for each dimension.
+    """
+    out = data_array
+    for dim, sl in slices.items():
+        out = out[dim, slice(*sl)].sum(dim=dim)
```
nvaytet, line 161: In my mind, the slider would just slice. It would then be up to the user to decide what to do after, possibly using another node to do the `sum` or `mean`.
So I would just return the sliced data.

Maybe we can then make a single `slice_dims` function that would work with both a single integer for slicing, as well as a range?

src/plopp/widgets/slice.py
```diff
@@ -88,14 +78,52 @@ def _plopp_observe_(self, callback: Callable, **kwargs):
...
+        super().__init__(da, dims, range=False)
+
+
+class RangeSliceWidget(_BaseSliceWidget):
+    """
+    Widgets containing a slider for each of the requested dimensions.
+    The widget uses the input data array to determine the range each slider should have.
+    Each slider also comes with a checkbox to toggle on and off the slider's continuous
+    update.
+
```
nvaytet, line 114:
```suggestion

    .. versionadded:: 24.04.0

```

nvaytet:
@MridulS how do you want to proceed with this?

MridulS:
Yes, I'll get this done later today/tomorrow :)

src/plopp/widgets/slice.py
MridulS: I think returning the slice and asking the user to add a new node to do the operation could be fine. PointCloud requires a single dimension and we need to do something with the returned slice to plot it out.

src/plopp/widgets/slice.py
```diff
@@ -88,14 +78,52 @@ def _plopp_observe_(self, callback: Callable, **kwargs):
...
+    Parameters
+    ----------
+    da:
+        The input data array.
+    dims:
+        The dimensions to make sliders for.
+    """
+
+    def __init__(self, da: sc.DataArray, dims: List[str]):
+        super().__init__(da, dims, range=True)
```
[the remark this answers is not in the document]
MridulS: I think it's nice but then should we also just add Float sliders too?

docs/gallery/scatter3d-with-slider.ipynb
```diff
@@ -100,6 +100,31 @@
...
+   "outputs": [],
+   "source": [
+    "# Use Plopp's widget to slice dimensions\n",
+    "slider = pw.RangeSliceWidget(da, dims=['time'])\n",
+    "slider_node = pp.widget_node(slider)\n",
+    "\n",
+    "# Sum over the selected range of time dimension\n",
+    "slice_node = pw.slice_dims(data_array=da, slices=slider_node)\n",
+    "\n",
+    "sum_slices = pw.operation_dim(data_array=slice_node, operation='sum', dim='time')\n",
```
nvaytet, line 117: I'm not sure I see the need for having a new function to perform operations.
How about something like
```Py
sum_slices = pp.Node(sc.sum, slice_node, dim='time')
```
?

src/plopp/widgets/slice.py
```diff
@@ -88,14 +78,54 @@ def _plopp_observe_(self, callback: Callable, **kwargs):
             self.controls[dim]['slider'].observe(callback, **kwargs)
 
     @property
-    def value(self) -> Dict[str, int]:
+    def value(self) -> Dict[str, int | tuple[int]]:
```
nvaytet, line 81: From which python version were these supported? I think we are saying Plopp sitll works with python 3.9. Do we need to use `Union` instead?

docs/gallery/scatter3d-with-slider.ipynb
MridulS: Yeah that makes more sense :)

nvaytet (approved):
Works nicely :+1:

src/plopp/widgets/__init__.py
```diff
@@ -7,7 +7,7 @@
 from .cut3d import Cut3dTool, TriCutTool
 from .drawing import DrawingTool, PointsTool
 from .linesave import LineSaveTool
-from .slice import SliceWidget, slice_dims
+from .slice import RangeSliceWidget, SliceWidget, slice_dims
 from .toolbar import Toolbar, make_toolbar_canvas2d, make_toolbar_canvas3d
 from .tools import ButtonTool, ColorTool, ToggleTool
```
nvaytet, line 13: @MridulS Can you fix? The CI is failing because we now need to add `RangeSliceWidget` to the `__all__` below.

PR #324, opened 2024-04-16
Small docs fixes

nvaytet:
- Add examples back in figure docstrings
- Deprecate `TriCutTool`

### DIFF
--- a/src/plopp/widgets/cut3d.py
+++ b/src/plopp/widgets/cut3d.py
@@ -225,6 +225,8 @@ class TriCutTool(ipw.HBox):
     A collection of :class:`Cut3dTool` to make spatial cuts in the X, Y, and Z
     directions on a three-dimensional scatter plot.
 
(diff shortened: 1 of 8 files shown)


src/plopp/widgets/cut3d.py
```diff
@@ -225,6 +225,8 @@ class TriCutTool(ipw.HBox):
     A collection of :class:`Cut3dTool` to make spatial cuts in the X, Y, and Z
     directions on a three-dimensional scatter plot.
 
+    .. deprecated:: v24.04.0
+
```
MridulS, line 229: Maybe a warning too like https://github.com/scipp/plopp/blob/c95e37596a8aa80ce2a30c35d965b5eb8268872b/src/plopp/graphics/figure.py#L11 ?

nvaytet: I now turned this into a decorator. What do you think?

PR #326, opened 2024-04-17
Protocols in typing

### LINKED ISSUE #318: Move protocols from `backends` to `core.typing`
Having it in `backends` doesn't really make sense and if moved to core it could be used more widely for type hinting (e.g. the `plot` function could return a `FigureLike` object.

nvaytet:
Fixes #318


jl-wynen (approved):
LGTM but I'm no type checker. How far are we from running mypy here?

nvaytet:
> LGTM but I'm no type checker. How far are we from running mypy here?

Very far, but maybe a good maintenance task to carry out?

PR #328, opened 2024-04-17
Fix some typing issues

jl-wynen:
The very first step to making mypy pass


src/plopp/core/__init__.py
```diff
@@ -7,3 +7,5 @@
 from .helpers import node, widget_node
```
nvaytet, line 7: Can you then remove the `flake8` noqas above? I think they were mostly related to importing something but not using it.

jl-wynen: Yes, good find!

nvaytet (approved):

PR #330, opened 2024-04-18
Add a couple of missing optional dependencies

nvaytet:
Note that `ipywidgets` was anyway included by `pythreejs`, but if one day we decide to move away from that, we would still need `ipywidgets` for the widgets.


MridulS (approved):

MridulS:
Can we make this non user editable?

PR #334, opened 2024-04-19
Fail on warnings during sphinx build

MridulS:
Let's test this out?

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -35,8 +35,8 @@ commands = pytest
 description = invoke sphinx-build to build the HTML docs
 deps = -r requirements/docs.txt
 allowlist_externals=find
-commands = python -m sphinx -j2 -v -b -W html -d {toxworkdir}/docs_doctrees docs html
-           python -m sphinx -j2 -v -b -W doctest -d {toxworkdir}/docs_doctrees docs html
+commands = python -m sphinx -W -j2 -v -b html -d {toxworkdir}/docs_doctrees docs html
+           python -m sphinx -W -j2 -v -b doctest -d {toxworkdir}/docs_doctrees docs html
            find html -type f -name "*.ipynb" -not -path "html/_sources/*" -delete
 
 [testenv:releasedocs]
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -35,8 +35,11 @@ commands = pytest
 description = invoke sphinx-build to build the HTML docs
 deps = -r requirements/docs.txt
 allowlist_externals=find
-commands = python -m sphinx -j2 -v -b html -d {toxworkdir}/docs_doctrees docs html
-           python -m sphinx -j2 -v -b doctest -d {toxworkdir}/docs_doctrees docs html
+# pydata_sphinx_theme raises a warning that it's not parallel-write-unsafe
+# https://github.com/pydata/pydata-sphinx-theme/issues/1643, removing -j2 for now
```
nvaytet, line 39: This is a good example. If we can't suppress the warning from `pydata-sphinx-theme`, maybe it's a good enough reason to not use `-W`?

MridulS: I mean it does catch an actual issue, it can cause problems with the sphinx build https://github.com/scipy/scipy/pull/16660#issuecomment-1888055872

nvaytet: True. We can try it for now and we can always remove it layer if it starts being annoying .

tox.ini
```diff
@@ -35,8 +35,11 @@ commands = pytest
 description = invoke sphinx-build to build the HTML docs
 deps = -r requirements/docs.txt
 allowlist_externals=find
-commands = python -m sphinx -j2 -v -b html -d {toxworkdir}/docs_doctrees docs html
-           python -m sphinx -j2 -v -b doctest -d {toxworkdir}/docs_doctrees docs html
+# pydata_sphinx_theme raises a warning that it's not parallel-write-unsafe
```
nvaytet, line 38:
```suggestion
# pydata_sphinx_theme raises a warning that it's not parallel-write-safe
```
?

nvaytet (approved):

PR #336, opened 2024-04-25
Start using the new bot and also bring ruff with copier

MridulS:


MridulS:
This will require some manual work to make ruff work first.

PR #341, opened 2024-05-17
Drop Python 3.9 and make Ruff happy

jl-wynen:
It may be easiest to review by going through the issues separately. `Apply pyupgrade fixes` is the lest important here.


nvaytet:
> It may be easiest to review by going through the issues separately. Apply pyupgrade fixes is the lest important here.

Did you mean going through the commits separately? ;-)
Also, is the pyupgrade fixes the most or the least important?

docs/gallery/rectangle-selection.ipynb
```diff
@@ -172,8 +176,23 @@
     "r._tool.click(50, 200)\n",
     "r._tool.click(200, 250)\n",
     "r._tool.click(30, 50)\n",
-    "r._tool.click(250, 170)\n",
-    "\n",
+    "r._tool.click(250, 170)"
+   ]
+  },
+  {
```
nvaytet, line 182: Same as above: annoying to have 2 hidden cells.

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -19,15 +19,13 @@ def _to_floats(x: np.ndarray) -> np.ndarray:
     return mdates.date2num(x) if np.issubdtype(x.dtype, np.datetime64) else x
 
 
-def _none_if_not_finite(x: Union[float, int, None]) -> Union[float, int, None]:
+def _none_if_not_finite(x: Union[float, None]) -> Union[float, int, None]:
```
nvaytet: Why did `int` go away here?

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -19,15 +19,13 @@ def _to_floats(x: np.ndarray) -> np.ndarray:
     return mdates.date2num(x) if np.issubdtype(x.dtype, np.datetime64) else x
 
 
-def _none_if_not_finite(x: Union[float, int, None]) -> Union[float, int, None]:
+def _none_if_not_finite(x: float | None) -> float | int | None:
```
nvaytet, line 22: Why was `int` removed here (and other places)?

.copier-answers.yml
```diff
@@ -1,9 +1,9 @@
 # Changes here will be overwritten by Copier; NEVER EDIT MANUALLY
-_commit: d91bf92
+_commit: 78693de
 _src_path: https://github.com/scipp/copier_template.git
 description: Visualization library for Scipp
 max_python: '3.12'
-min_python: '3.9'
+min_python: '3.10'
```
nvaytet, line 6: I think you also need to bump the python version for the scipp wheel in the `nightly.txt` requirement.

docs/gallery/interactive-masking.ipynb
```diff
@@ -320,7 +332,8 @@
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3"
+   "pygments_lexer": "ipython3",
+   "version": "3.10.14"
```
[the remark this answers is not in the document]
jl-wynen: You could open an issue to ask them to add this feature.

docs/gallery/interactive-masking.ipynb
```diff
@@ -211,8 +208,23 @@
     "r._tool.click(50, 200)\n",
     "r._tool.click(200, 250)\n",
     "r._tool.click(30, 50)\n",
-    "r._tool.click(250, 170)\n",
-    "\n",
+    "r._tool.click(250, 170)"
+   ]
+  },
+  {
```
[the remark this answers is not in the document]
jl-wynen: Yes, because of the import. I did not see that the cells are hidden. Merged them again.

src/plopp/backends/matplotlib/canvas.py
jl-wynen: Because `int | float` is the same as `float` to mypy. This is technically not clean because `int` is not a subclass of `float`. But mypy treats it as such. And ruff pointed that out.

jl-wynen:
I don't understand why the docs build fails. It seems like it wants to access `StatigFig.toolbar` but that does not exist. This broke in the dependency-update commit. @nvaytet do you have any idea what might be going on.

src/plopp/backends/matplotlib/canvas.py
nvaytet: > But mypy treats it as such

I don't completely understand why we want to make mypy happy when it's asking for something which is not technically correct. Also, it seems weird that the `int` was removed from the input type, but not from the return type?

nvaytet:
> It seems like it wants to access StatigFig.toolbar but that does not exist.

This would mean it's trying to make a static figure when it should be making an interactive one...
To make interactive figs in the docs, we set a variable in the `conf.py`:
```Py
nbsphinx_execute_arguments = [
    "--Session.metadata=scipp_sphinx_build=True",
]
```
Could it be a change in sphinx that broke this?

Alternatively a change in `matplotlib` or `ipympl`?

src/plopp/backends/matplotlib/canvas.py
MridulS: > I don't completely understand why we want to make mypy happy when it's asking for something which is not technically correct. 

We have to make one of the type checkers happy, otherwise we aren't really testing out the type annotations. The reasoning is given in this PEP https://peps.python.org/pep-0484/#the-numeric-tower. We either open issues upstream or live with the fact `float` includes `int` with python type hints 🤷

nvaytet: Ok for the tower, but it still doesn't answer why the return type still has `float | int`?

jl-wynen: I don't know. I can't think of a case where it would make a difference.
Technically, PEP484 only specifies this behaviour for arguments, not return types. But I tried it out with mypy and it seems to behave the same for the return type.

jl-wynen:
Allegedly, there is no difference between the two when running in Jupyter: https://github.com/matplotlib/ipympl/issues/529
Is this a bug upstream?

nvaytet:
It's weird, when you are in a notebook, and you ask for the backend with `matplotlib.get_backend()`, you get `'module://ipympl.backend_nbagg'` with both `%matplotlib widget` and `%matplotlib ipympl`.
However, when building sphinx docs, if you print `matplotlib.get_backend()` it returns just `widget` with `%matplotlib widget` in the notebook (I am guessing it will print `ipympl` with `%matplotlib ipympl` but I have not actually checked that).

In any case, I pushed a fix which seems to work.

jl-wynen:
Are you happy with the rest?

nvaytet (approved):

PR #344, opened 2024-05-22
Allow to remove nodes with children

nvaytet:
Until now, we we preventing the removal of nodes that had children, to avoid making isolated nodes that were unreachable in a graph.
However, especially with interactive tools, there are some cases where nodes are dynamically added and removed, either on a view or on other nodes.

There can also be nodes that take in an arbitrary number of arguments. The following example makes sense and should be allowed:
```Py
a = Node(lambda: 5)
b = Node(lambda: 3)
c = Node(lambda: 11)
d = Node(lambda *args: sum(args), a, b, c)
assert d() == 19
a.remove()
assert d() == 14
b.remove()
assert d() == 11
c.remove()
assert d() == 0
```

### DIFF
--- a/src/plopp/core/node.py
+++ b/src/plopp/core/node.py
@@ -105,10 +105,14 @@ class Node:
         The operation fails is the node has children, as removing it would leave the
         graph in an ill-defined state.
         """
-        if self.children:
-            raise RuntimeError(
-                f"Cannot delete node because it has children {self.children}."
-            )
+        for child in self.children:
+            if self in child.parents:
+                child.parents.remove(self)
+            child.kwparents = {
+                key: parent for key, parent in child.kwparents.items() if parent != self
+            }
+            child._data = None
+        self.children.clear()
         for view in self.views:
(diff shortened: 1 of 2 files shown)


src/plopp/core/node.py
```diff
@@ -105,17 +105,22 @@ def remove(self) -> None:
         The operation fails is the node has children, as removing it would leave the
         graph in an ill-defined state.
         """
-        if self.children:
-            raise RuntimeError(
-                f"Cannot delete node because it has children {self.children}."
-            )
+        for child in self.children:
+            if self in child.parents:
```
YooSunYoung, line 109: Why is this check needed?

nvaytet: Because we don't know if `self` is in the `parents` or the `kwparents`.

YooSunYoung: Ah..! right. You handle the `kwparents` below. Thank you!

YooSunYoung (approved):

PR #347, opened 2024-05-31
Remove old unused nodes that were left over when removing shapes

nvaytet:


src/plopp/widgets/drawing.py
```diff
@@ -106,6 +106,7 @@ def remove_node(self, artist):
             self._destination.canvas.draw()
         output_node.remove()
         draw_node.remove()
+        del self._draw_nodes[nodeid]
```
jl-wynen, line 109: Could you just use `self._draw_nodes.pop(nodeid)` above instead of `__getitem__`?

jl-wynen (approved):

PR #349, opened 2024-06-17
The `_plopp_mask` on line objects needs to be updated after data update

nvaytet:
`_plopp_mask` is a special property we have added on mpl Line object and is used to identify the points to consider when trying to autoscale the axes.
This was discovered when resizing rectangles in the [rectangle selection](https://scipp.github.io/plopp/gallery/rectangle-selection.html) notebook.

Not sure how to unit-test, as `pick` events currently [cannot be simulated](https://stackoverflow.com/questions/74771867/simulate-click-event-in-matplotlib-that-triggers-a-pick-event).
Suggestions welcome!

### DIFF
--- a/src/plopp/backends/plotly/line.py
+++ b/src/plopp/backends/plotly/line.py
@@ -192,11 +192,13 @@ class Line:
         """
         self._data = new_values
         new_values = make_line_data(data=self._data, dim=self._dim)
+        line_mask = ~np.isnan(new_values['mask']['y'])
 
         with self._fig.batch_update():
             self._line.update(
                 {'x': new_values['values']['x'], 'y': new_values['values']['y']}
             )
+            self._line._plopp_mask = line_mask
 
             if (self._error is not None) and (new_values['stddevs'] is not None):
                 self._error.update(
@@ -205,6 +207,9 @@ class Line:
                         'y': new_values['stddevs']['y'],
                         'error_y': {'array': new_values['stddevs']['e']},
                     }
+                )
+                self._error._plopp_mask = (
+                    line_mask[1:] if new_values["hist"] else line_mask
                 )
 
             if new_values['mask']['visible']:
--- a/src/plopp/backends/matplotlib/line.py
+++ b/src/plopp/backends/matplotlib/line.py
@@ -180,6 +180,9 @@ class Line:
         self._mask.set_data(new_values['mask']['x'], new_values['mask']['y'])
         self._mask.set_visible(new_values['mask']['visible'])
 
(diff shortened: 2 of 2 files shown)


src/plopp/backends/plotly/line.py
```diff
@@ -192,11 +192,13 @@ def update(self, new_values: sc.DataArray):
         """
         self._data = new_values
         new_values = make_line_data(data=self._data, dim=self._dim)
+        line_mask = ~np.isnan(new_values['mask']['y'])
```
SimonHeybrock, line 195: It is unclear why this is needed, and not any of the other things the `_make_line` method above does. The code is duplicate. What if one of them is changed? `__init__` seems to be calling `_make_line` on `_make_line_data`... maybe something can/should be moved around?

nvaytet: The `_plopp_mask` is needed because the canvas is in charge of autoscaling the axes (this was just a choice, it does not necessarily have be this way). However, it does not know about the Plopp lines or images that are shown onto it, it just handles the axes.

To perform auto-scaling, it looks at the matplotlib artists it holds, but needs to handle lines made by plopp in a special way , as they come with an (most of the time) invisible line of masks (which is why we can't just let the plotting library do the work, unfortunately, also it is bad at handling log axes, and we do better there).

All this feels (and has felt for a long time) a bit weird: the canvas doesn't know about the lines or images, but actually it does because it has to do the auto-scaling. Maybe the auto-scaling should be done by someone else who does have knowledge of the artists. Maybe each Line and Image should provide what its bounding box is (taking into account its own masks etc), and the auto-scaling should just be a simple loop over artists and merge all the bounding boxes.

> `__init__` seems to be calling `_make_line` on `_make_line_data`... maybe something can/should be moved around?

Yes, I don't like this either, and can't really remember why the `_make_line` function exists.

I have plans to rework all the internals of how the plotting libraries hook in as backends, as I feel there is an awful lot of code that is duplicated between different plot types etc (the idea is kind of drafted here: https://github.com/scipp/plopp/compare/mesh-3d...simpler-backends).

All this means that it may take a while before all these changes make it in. The issue I found here is a bug that makes the rectangle selection example in the gallery very unusable.
Should we just get these changes in, make a patch release and then work on the large changes?

SimonHeybrock (approved):

src/plopp/backends/plotly/line.py
SimonHeybrock: > Should we just get these changes in, make a patch release and then work on the large changes?

Let us not make large changes, as long as it is working and not making other ongoing changes a hassle. Refactor only when there is a good reason.

PR #353, opened 2024-08-02
Rewrite backends implementation to make futher additions easier

nvaytet:
This PR's goal is to reduce the amount of code one has to add if we wish to add a new plot type.
For example, if we want to add a new contour plot in the matplotlib backend, currently we need to:

1. Create a `backends/matplotlib/contour.py` file which actually implements the contour plot in matplotlib
2. Add the `contour` to the `MatplotlibBackend`: https://github.com/scipp/plopp/blob/main/src/plopp/backends/matplotlib/__init__.py#L18
3. Add the `contour` to the `BackendManager`: https://github.com/scipp/plopp/blob/main/src/plopp/backends/manager.py#L7
4. Create a `graphics/contourview.py` that contains a `ContourView` and a `contourfigure` (lots of code and repetition of arguments/docstrings)
5. Create a top-level `plotting/contour.py` function available for the user.

In this PR, instead of hard-coding everything in the `MatplotlibBackend` and the `BackendManager`, we dynamically import the classes as needed. **Reviewer: have a look at this part to see if there could be issues with the approach.**
Code is here: https://github.com/scipp/plopp/blob/backends-rewrite/src/plopp/backends/__init__.py#L10

This means that to add a new contour plot, only steps 1 and 5 remain (plus we need to add one small function to https://github.com/scipp/plopp/blob/backends-rewrite/src/plopp/graphics/figures.py)

~This is an early draft, just to get some feedback if the approach is ok, or not.~
~Most plots should work: line, image, scatter, scatter3d (note that you now need `cbar=True`)...~

**Notable changes:**

- `LineView`, `ImageView`, `ScatterView`, and `Scatter3dView` have been removed, only `GraphicalView` remains.
- Autoscaling was moved from the canvas to the `GraphicalView`, and each artist provides its own bounding box.
- The deprecated `figure1d`, `figure2d`, `figure3d` were removed (changes would have been required to get them to work)
- The deprecated `Cut3d` tool was removed (changes would have been required to get it to work).
- The `pixel_size` argument for the `scatter3d` plot was deprecated in favour of `size` which is consistent with the 2d `scatter` plot.
- The `scatter3d` plot now requires `cbar=True` argument to show a colorbar.
- Many tests have changed/moved: many would have needed refactoring as they contained implementation details. I am now trying to test behaviours rather than implementation details.


jl-wynen (commented):
The general mechanism seems fine to me. But it looks like there are unrelated changes in this PR. Is this correct? if so, can you remove them to make it easier to see the impact of the refactor?

src/plopp/backends/matplotlib/canvas.py
```diff
@@ -90,9 +73,10 @@ def __init__(
         figsize: tuple[float, float] | None = None,
         title: str | None = None,
         grid: bool = False,
-        vmin: sc.Variable | float = None,
-        vmax: sc.Variable | float = None,
-        autoscale: Literal['auto', 'grow'] = 'auto',
+        # bbox: BoundingBox | None = None,
+        user_vmin: sc.Variable | float = None,
+        user_vmax: sc.Variable | float = None,
+        # autoscale: Literal['auto', 'grow'] = 'auto',
```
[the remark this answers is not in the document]
nvaytet: While it does seems unrelated, but it was actually necessary to move the autoscaling mechanism from the canvas to the `GraphicalView`, to unify the behaviors between the 2d and 3d canvas, because everything now goes through the same `GraphicalView` (there is no more `Scatter3dView`).

I'll clean it up and then can do a walkthrough to help with the review process.

src/plopp/backends/matplotlib/figure.py
```diff
@@ -1,16 +1,24 @@
...
-# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
 
 from ...graphics import BaseFig
+from .utils import fig_to_bytes
 
+try:
+    from ipywidgets import VBox
+except ImportError:
+    VBox = object
```
jl-wynen, line 10: Is there a check elsewhere that raises a user-friendly error message? If not, this code here would lead to `AttributeError`s without further explanation if `ipywidgets` is not installed.

src/plopp/backends/matplotlib/tiled.py
```diff
@@ -9,8 +9,25 @@
...
+        dims=fig.view._dims, canvas_maker=Canvas, artist_maker=fig.view._artist_maker
+    )
+    out = fig.__class__(
+        fig._view_maker,
+        *fig._args,
+        **args,
+    )
+    for prop in ('xrange', 'yrange', 'xscale', 'yscale', 'title', 'grid'):
+        setattr(out.canvas, prop, getattr(fig.canvas, prop))
+    return out
```
jl-wynen, line 30: This function accesses a bunch of protected attributes. Can you move it into `MplBaseFig`? Or add `MplBaseFig.copy` and use that here?

src/plopp/backends/matplotlib/image.py
```diff
@@ -75,13 +77,15 @@ def __init__(
...
         data: sc.DataArray,
+        artist_number: int = 0,
         shading: str = 'auto',
         rasterized: bool = True,
         **kwargs,
     ):
         self._canvas = canvas
         self._ax = self._canvas.ax
         self._data = data
+        self._artist_number = artist_number
```
jl-wynen, line 88: Is this used anywhere?

src/plopp/backends/matplotlib/line.py
```diff
@@ -57,110 +68,77 @@ def __init__(self, canvas: Canvas, data: sc.DataArray, number: int = 0, **kwargs
...
-                yerr=data['stddevs']['e'],
+                line_data['stddevs']['x'],
+                line_data['stddevs']['y'],
+                yerr=line_data['stddevs']['e'],
                 color=self._line.get_color(),
                 zorder=10,
                 fmt="none",
             )
-            # Set the selection mask on the line collection that makes the segments
-            self._error[2][0]._plopp_mask = line_mask[1:] if data["hist"] else line_mask
```
jl-wynen, line 162: Why has this been removed?

src/plopp/graphics/graphicalview.py
```diff
@@ -15,22 +31,112 @@ class GraphicalView(View):
...
         new = dict(*args, **kwargs)
         for key, new_values in new.items():
-            if new_values.ndim != self._ndim:
-                raise ValueError(
-                    f"Expected {self._ndim} dimension(s), but got {new_values.ndim}."
-                )
+            # if new_values.ndim != self._ndim:
+            #     raise ValueError(
+            #         f"Expected {self._ndim} dimension(s), but got {new_values.ndim}."
+            #     )
```
jl-wynen, line 139: Remove or re-enable?

src/plopp/graphics/tiled.py
```diff
@@ -52,4 +52,5 @@ def tiled(nrows: int, ncols: int, **kwargs):
       >>> tiled[0, 2] = da2.plot()
 
     """
-    return backends.tiled(nrows=nrows, ncols=ncols, **kwargs)
+    return backends.get(group='2d', name='tiled')(nrows=nrows, ncols=ncols, **kwargs)
+    # return backends.tiled(nrows=nrows, ncols=ncols, **kwargs)
```
jl-wynen, line 56:
```suggestion
```

tests/graphics/canvas_test.py
```diff
@@ -0,0 +1,217 @@
...
+    backends['2d'] = backend
+    return linefigure(*args, **kwargs)
+
+
+CASES = [
+    FigureAndData(partial(backend_linefigure, backend='matplotlib'), data1d),
+    FigureAndData(partial(backend_linefigure, backend='plotly'), data1d),
+    FigureAndData(partial(imagefigure, cbar=True), data2d),
+    FigureAndData(partial(scatterfigure, x='x', y='y', cbar=True), scatter),
+    FigureAndData(partial(scatter3dfigure, x='x', y='y', z='z', cbar=True), scatter),
```
jl-wynen, line 30: This looks suspicious. `backend_linefigure` sets the backend. But the other functions don't. Why is that. (I know that there is a fixture that resets the backend.)

src/plopp/backends/matplotlib/line.py
nvaytet: Good question. This is related to moving the autoscaling from the `Canvas` to the `GraphicalView`.

Before:
The canvas would identify all the graphics it had (looking at matplotlib artists, not our objects) and try to compute the global min/max limits for all artists. Because we wanted to ignore the masked points when searching for the limits, I had to add the mask on the matplotlib object so it could be read at the canvas level.

Now:
The GraphicalView just loops over its Plopp artists, and asks each one of them for their bounding box.
It computes a global bounding box from all of them.
The `Line` artists knows its own mask, so it just returns the corretc bounding box, the View never needs to know about this.

tests/graphics/canvas_test.py
nvaytet: Yes, it is relying on the fixture that resets the backends.

I basically wanted to run the same tests with all possible cases, including plotly.
Didn't really know how else to do it. I thought about trying to use a fixture, as done [here](https://github.com/scipp/plopp/blob/main/tests/backends/plotly/conftest.py#L9), but wasn't sure how to.

Do you have a suggestion?

nvaytet:
Just noticed an issue with the bounding box when making 2d figures with `matplotlib widget`: the edges of the figure are missing.

src/plopp/backends/matplotlib/image.py
nvaytet: Not in the image, but the other artists (scatter, line) have one to keep track of the color to use.
I will ignore it in a better way.

src/plopp/backends/matplotlib/figure.py
nvaytet: No there is no check elsewhere, but I am not sure you can enable `%matplotlib widget` without having `ipympl` installed, and that itself requires `ipywidgets`?
I will double check.

tests/graphics/canvas_test.py
jl-wynen: I think it is ok for tests to select a backend. I was mostly surprised that there is only one function that does it while the other functions just use the global default backend.

src/plopp/backends/matplotlib/figure.py
jl-wynen: So does this code only use `VBox` when widgets are enabled? Not when a user builds custom widgets using Plopp's API?

nvaytet: Result of the check: if you try to do `%matplotlib widget` in a notebook without ipywidgets installed, you get
```
RuntimeError: 'widget is not a recognised GUI loop or backend name
```

> So does this code only use VBox when widgets are enabled

Yes, the `VBox` here should only be used to create a figure when `%matplotlib widget` is enabled. If it isn't, a `StaticFigure` is created instead.

> Not when a user builds custom widgets using Plopp's API?

If a user is building custom widgets, I'm guessing they must have `ipywidgets` installed?

The try/except here is to ensure that `ipywidgets` is not a hard dependency when we import, because `InteractiveFigure` inherits from `VBox`.
Maybe there is a better way to do this?

jl-wynen: I'm sure there are more complicated ways to do it and have a custom error message. But I think it's fine. I was just worried about people getting bad error messages from Plopp. But if they have to manually import ipywidgets anyway when building custom widgets, that's fine.

src/plopp/backends/matplotlib/tiled.py
nvaytet: Done.

src/plopp/graphics/graphicalview.py
nvaytet: So I had disabled this because it no longer made so much sense for the `GraphicalView` to know about the dimensionality of the data. If you supplied data with the wrong number of dimensions, the figure update would still fail, but the error messages would be quite unclear.

I have worked around this in 2 ways:
- In the `GraphicalView`, I check if the supplied data has the coordinates it needs to have, and if not provide info (instead of just a key error)
- In the artists themselves (`Line`, `Image`, `Scatter`, and `Scatter3d`), I perform the dimension check as the artists know what kind of data they are expecting to get (this was already done in the `Scatter3d` and I copied the mechanism over to the others).

src/plopp/graphics/graphicalview.py
```diff
@@ -195,6 +201,8 @@ def update(self, *args, **kwargs) -> None:
 
             self.artists[key].update(new_values=new_values)
 
+            print('self._dims', self._dims)
```
jl-wynen:
```suggestion
```

Update ruff to make it catch this.

jl-wynen (changes requested):

jl-wynen (approved):

PR #364, opened 2024-09-05
Add example of streaming data to a figure in the gallery

nvaytet:
Using the [Play widget](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html#play-animation-widget) from ipywidgets, we stream updates to a plot while being able to pan and zoom.

Screenshot at 2024-09-05 15-37-59


nvaytet:
Merging as it's only a gallery notebook, it doesn't break anything in Plopp.

PR #367, opened 2024-09-11
Fix clipped figures in horizontal layout

nvaytet:
Before:
Screenshot at 2024-09-11 10-12-26

After:
Screenshot at 2024-09-11 10-11-54

I also made sure that the 100% layout for the 3d figures still makes sure the toolbar doesn't get hidden when the canvas gets too large.

### DIFF
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -134,19 +134,15 @@ class Canvas:
         return Image(value=fig_to_bytes(self.fig), format='png')
 
     def to_widget(self):
-        from ipywidgets import Layout, VBox
+        from ipywidgets import VBox
 
         if self.is_widget() and not is_sphinx_build():
             self.fig.tight_layout()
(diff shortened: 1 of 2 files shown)


src/plopp/backends/matplotlib/canvas.py
```diff
@@ -134,19 +134,15 @@ def to_image(self):
...
         if self.is_widget() and not is_sphinx_build():
             self.fig.tight_layout()
             # The Matplotlib canvas tries to fill the entire width of the output cell,
             # which can add unnecessary whitespace between it and other widgets. To
             # prevent this, we wrap the canvas in a VBox, which seems to help.
-            widget = VBox([self.fig.canvas])
-        else:
-            widget = self.to_image()
-        # The max_width is set to prevent overflow, see gh-169
-        widget.layout = Layout(max_width='80%', overflow='auto')
```
nvaytet, line 148: This was here because of https://github.com/scipp/plopp/pull/319#issuecomment-2037421653.
But I actually tried to make a very large figure and it didn't make a difference, so removing it since it's not doing anything.

PR #368, opened 2024-09-11
BUILD: kaleido should be python-kaleido for conda build

MridulS:
We should probably change this upstream in the copier template, but I guess we can unblock the release build of plopp for now.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -35,7 +35,7 @@ test:
 
   {# Conda does not allow spaces between package name and version, so remove them #}
   {% for package in test_dependencies %}
-    - {% if package == "graphviz" %}python-graphviz{% else %}{{ package|replace(" ", "") }}{% endif %}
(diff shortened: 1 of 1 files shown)


conda/meta.yaml
```diff
@@ -35,7 +35,7 @@ test:
 
   {# Conda does not allow spaces between package name and version, so remove them #}
   {% for package in test_dependencies %}
-    - {% if package == "graphviz" %}python-graphviz{% else %}{{ package|replace(" ", "") }}{% endif %}
+    - {% if package == "graphviz" %}python-graphviz{% elif package == "kaleido" %}python-kaleido{% else %}{{ package|replace(" ", "") }}{% endif %}
```
nvaytet, line 38: If we get more examples of this, we could do something like prepend `python-` to the package name if `name in ["graphviz", "kaleido", ...]`, but this is fine for now.

nvaytet (approved):

PR #370, opened 2024-09-12
Add data array name on vertical axis of 1d plots

### LINKED ISSUE #356: Add ylabel on 1d plots if only 1 line is plotted?
Currently only the unit is shown on the ylabel of a 1d plot. If the data has a name, we could display it as long as only one curve is on the graph.

This is already the case for the label on the color bar for 2D plots.

nvaytet:
This adds the name of the data array if defined on the vertical axis of 1d plots, as long as there is only one data array plotted. If there is more than one curve, only the unit is on the axis, and the legend is shown instead.

Examples:
```Py
import plopp as pp

a = pp.data.data1d()
a.name = "Velocity"
a.plot()
```
Screenshot at 2024-09-12 15-00-59

```Py
import plopp as pp

a = pp.data.data1d()
b = a * 2
pp.plot({'a': a, 'b': b})
```
Screenshot at 2024-09-12 15-01-36

Fixes #356


src/plopp/backends/matplotlib/line.py
```diff
@@ -187,7 +182,7 @@ def remove(self):
         self._mask.remove()
         if self._error is not None:
             self._error.remove()
-        self.update_legend()
+        # self.update_legend()
```
jl-wynen, line 185:
```suggestion
```

jl-wynen (approved):
Looks good after a tiny fix.

PR #372, opened 2024-09-13
Raise error when two coords with same underlying dimension are used

### LINKED ISSUE #360: coord arg accepts multiple coords for same dimension
The `coords` argument accepts multiple coords for the same dimension and uses the last one in the list:
Screenshot_20240904_100350

This is surprising to me. Should it instead raise an error?

nvaytet:
Fixes #360

### DIFF
--- a/src/plopp/plotting/common.py
+++ b/src/plopp/plotting/common.py
@@ -198,7 +198,7 @@ def preprocess(
             underlying = out.coords[dim].dims[-1]
             if underlying in renamed_dims:
                 raise ValueError(
-                    "coords: Cannot use the more than one coordinate associated with"
+                    "coords: Cannot use more than one coordinate associated with"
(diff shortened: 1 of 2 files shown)


src/plopp/plotting/common.py
```diff
@@ -196,6 +196,11 @@ def preprocess(
             coords = [coords]
         for dim in coords:
             underlying = out.coords[dim].dims[-1]
+            if underlying in renamed_dims:
+                raise ValueError(
+                    "coords: Cannot use more than one coordinate associated with"
```
jl-wynen, line 201:
```suggestion
                    "coords: Cannot use more than one coordinate associated with "
```

jl-wynen (changes requested):

jl-wynen (approved):

PR #376, opened 2024-09-17
Add mesh 3d

nvaytet:
This adds a plot that can take in a list of vertices and faces and construct a 3D mesh.

This opens the way for representing full instrument geometries in 3D.

Notes:

For now we only support vertex colors, we could add face colors in the future.
This also supports showing only a single mesh at a time, unlike the scatter3d plot who can take in a dict of data arrays.

Supersedes #346


nvaytet:
@g5t would be great if you could also give this a go, to see if it does what you want.

src/plopp/backends/pythreejs/mesh3d.py
```diff
@@ -0,0 +1,179 @@
...
+        self,
+        *,
+        canvas: Canvas,
+        data: sc.DataArray,
+        color: str | None = None,
+        opacity: float = 1,
+        edgecolor: str | None = None,
+        artist_number: int = 0,
+    ):
+        import pythreejs as p3
```
jl-wynen, line 51: What if pythreejs is not installed or python is not running in Jupyter? Does the backend manager raise an appropriate exception or does this need to happen here?

src/plopp/backends/pythreejs/mesh3d.py
```diff
@@ -0,0 +1,179 @@
...
+
+        attributes = {
+            'position': p3.BufferAttribute(
+                array=np.array(
+                    [
+                        self._data.coords["x"].values.astype('float32', copy=False),
+                        self._data.coords["y"].values.astype('float32', copy=False),
+                        self._data.coords["z"].values.astype('float32', copy=False),
+                    ]
+                ).T
```
jl-wynen, line 73: Don't know what size of mesh this can or should support. But you could avoid extra copies here by not taking x,y,z apart into separate variables. Instead, you could use
```python
p3.BufferAttribute(positions.values.astype('float32', copy=False))
```
(`copy=False` doesn't do anything here because the positions vector is always `float64`.)

src/plopp/backends/pythreejs/mesh3d.py
```diff
@@ -0,0 +1,179 @@
...
+        Parameters
+        ----------
+        new_values:
+            New data to update the mesh values from.
+        """
+        self._data = new_values
+        # TODO: for now we only update the colors of the mesh. Updating the positions
+        # of the vertices is doable but is made more complicated by the edges geometry,
+        # whose positions cannot just be updated. A new geometry and edge lines would
+        # have to be created, the old one removed from the scene and the new one added.
```
jl-wynen, line 135: It doesn't looks like `update` changes the colours to me. That is what `set_colors` does. So could `update` simply raise an exception instead.
As it stands, this could hide errors.

docs/user-guide/plot-types/mesh3d-plot.ipynb
```diff
@@ -0,0 +1,117 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2",
+   "metadata": {},
+   "source": [
+    "## Loading mesh data\n",
+    "\n",
+    "We load a file containing the data to construct the [Utah teapot](https://en.wikipedia.org/wiki/Utah_teapot)."
```
jl-wynen, line 33: Can you explain the data format? It is not obvious to me how to encode vertices and faces.

src/plopp/backends/pythreejs/mesh3d.py
nvaytet: No idea. It's the same in `scatter3d`.
I don't think it's an issue not running in Jupyter, as the unit tests pass fine.

If it's not installed, you will just get `ImportError` on `pythreejs`?

src/plopp/backends/pythreejs/mesh3d.py
nvaytet: I split up the coordinates because I wanted the [figure](https://github.com/scipp/plopp/pull/376/files#diff-b9679c47242db4cccfbd303f917abb03780e549cfaf786c5f0bdbcb8682da0f3R81) to behave a bit like `scatter3d`, with checking the units of the x,y,z coordinates. But maybe I can make something that just takes in a `vertices` coordinate...

src/plopp/backends/pythreejs/mesh3d.py
nvaytet: It's true that the colors are not changed. However, the values of the `._data` should be updated, because this is what the `colormapper` uses to determine the colorbar range. So it should not raise, but I will update the comment.

src/plopp/backends/pythreejs/mesh3d.py
jl-wynen: > If it's not installed, you will just get ImportError on pythreejs?

I would assume so. Maybe this is something that the backend manager can catch so we don't have to handle it in everywhere. But that is unrelated to this PR.

jl-wynen (approved):

PR #378, opened 2024-09-24
Improve test coverage by parametrizing tests

nvaytet:
Parametrize many tests over the different backends:

- MPL-static
- MPL-interactive
- plotly
- pythreejs

Total number of tests went up from 620 to 912 :-)


src/plopp/backends/matplotlib/canvas.py
```diff
@@ -157,7 +157,10 @@ def to_widget(self):
         from ipywidgets import VBox
 
         if self.is_widget() and not is_sphinx_build():
-            self.fig.tight_layout()
+            try:
```
nvaytet, line 160: This came up when trying to make tiled plots with the `ipympl` backend.

pyproject.toml
```diff
@@ -81,6 +81,8 @@ filterwarnings = [
   'ignore:setDaemon\(\) is deprecated, set the daemon attribute instead:DeprecationWarning',
   'ignore:TriCutTool is deprecated::',
   'ignore:Cut3dTool is deprecated::',
+  'ignore:This figure includes Axes that are not compatible with tight_layout:UserWarning',
+  'ignore:The figure layout has changed to tight:UserWarning',
```
jl-wynen, line 85: Why do these need to be ignored? It seems to me like Plopp should not raise warnings like this unless they are caused by the user.

nvaytet: This is mostly to do with the tiled plots, where the axes handling is a little funny. Maybe it's better to catch the warnings in the code, so that we are sure this is the place they are coming from?

jl-wynen: > Maybe it's better to catch the warnings in the code, so that we are sure this is the place they are coming from?

:+1:

jl-wynen (approved):

PR #379, opened 2024-09-24
Restore toolbar for tiled plots

nvaytet:
Fixes 2 cases:

1. When making `tiled` plots using the `%matplotlib widget` backend, the toolbar would disappear.

```Py
import plopp as pp
%matplotlib widget

f = pp.data.data1d().plot()
f / f
```

2. Toolbar would also be absent when making figures using custom axes, e.g.:

```Py
import plopp as pp
%matplotlib widget

fig, ax = plt.subplots(2, 1)
f1 = pp.data.data1d().plot(ax=ax[0])
f2 = pp.data.data2d().plot(ax=ax[1])
fig.canvas
```

The fix is to not alter the visibility of the toolbar if axes come from the outside world.

Note that this does mean that if you do something like
```Py
fig, ax = plt.subplots(2, 1)
f = pp.data.data1d().plot(ax=ax[0])
f
```
you will see 'double' toolbar, but usually if one supplies the `ax` argument, it implies that we want to use the matplotlib figure instead of the Plopp figure.

### DIFF
--- a/src/plopp/backends/matplotlib/canvas.py
+++ b/src/plopp/backends/matplotlib/canvas.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
+import warnings
 from typing import Literal
 
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


nvaytet:
Changing to draft as #378 needs to be merged first.

MridulS (approved):

PR #380, opened 2024-09-25
Fix showing figures from terminal

### LINKED ISSUE #342: Showing figures from the terminal does not work
Example:

Open a terminal and start `ipython`
```Py
import plopp as pp
da = pp.data.data1d()
fig = da.plot()
fig.canvas.show()
```
One would expect a Qt window to pop up with the plot inside it, but instead we get the error:
```
AttributeError: Figure.show works only for figures managed by pyplot, normally created by pyplot.figure()
```

I am thinking it would be something to do with this change: https://github.com/scipp/plopp/pull/290/files#diff-34ac43cb9dac33f569731376decbd9b286cb0985da4c8de3443bca5551377d00R37

nvaytet:
Fixes #342 

I also removed the top-level `plopp.show()` function which probably did not work and I don't think was ever used.

### DIFF
--- a/src/plopp/backends/matplotlib/utils.py
+++ b/src/plopp/backends/matplotlib/utils.py
@@ -27,10 +27,16 @@ def fig_to_bytes(fig: plt.Figure, form: Literal['png', 'svg'] = 'png') -> bytes:
 
 def is_interactive_backend() -> bool:
     """
-    Return ``True`` if the current backend used by Matplotlib is the widget/ipympl
-    backend.
+    Return ``True`` if the current backend used by Matplotlib creates interactive
+    figures. See
+    https://matplotlib.org/stable/users/explain/figure/backends.html#the-builtin-backends
+    for a list of backends.
     """
-    return "inline" not in mpl.get_backend()
+    backend = mpl.get_backend().lower()
+    return any(
+        b in backend
+        for b in ('qt', 'ipympl', 'gtk', 'tk', 'wx', 'nbagg', 'web', 'macosx', 'widget')
+    )
 
 
 def make_figure(*args, **kwargs) -> plt.Figure:
@@ -51,7 +57,10 @@ def make_figure(*args, **kwargs) -> plt.Figure:
     if is_interactive_backend():
         # Create a manager for the figure, which makes it interactive, as well as
         # making it possible to show the figure from the terminal.
-        plt._backend_mod.new_figure_manager_given_figure(1, fig)
+        try:
+            plt._backend_mod.new_figure_manager_given_figure(1, fig)
+        except AttributeError:
+            pass
     return fig
 
 
--- a/src/plopp/backends/matplotlib/utils.py
+++ b/src/plopp/backends/matplotlib/utils.py
@@ -5,10 +5,12 @@ from io import BytesIO
 from typing import Literal
 
 import matplotlib as mpl
-from matplotlib.pyplot import Figure, _get_backend_mod
+
+# from matplotlib.pyplot import Figure, _backend_mod, _get_backend_mod
+import matplotlib.pyplot as plt
 
 
-def fig_to_bytes(fig: Figure, form: Literal['png', 'svg'] = 'png') -> bytes:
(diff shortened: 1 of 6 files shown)


src/plopp/backends/matplotlib/utils.py
```diff
@@ -27,14 +27,19 @@ def fig_to_bytes(fig: Figure, form: Literal['png', 'svg'] = 'png') -> bytes:
...
+    figures. See
+    https://matplotlib.org/stable/users/explain/figure/backends.html#the-builtin-backends
+    for a list of backends.
     """
-    backend = mpl.get_backend()
-    return any(x in backend for x in ("ipympl", "widget"))
+    backend = mpl.get_backend().lower()
+    return any(
+        b in backend
+        for b in ('qt', 'ipympl', 'gtk', 'tk', 'wx', 'nbagg', 'web', 'macosx', 'widget')
```
MridulS, line 38: Not sure if this is any prettier but this can be accessed via
```
matplotlib.backends.backend_registry.list_builtin(matplotlib.backends.BackendFilter.INTERACTIVE)
```

MridulS: The widget one is missing from the list.

src/plopp/backends/matplotlib/utils.py
```diff
@@ -45,15 +50,18 @@ def make_figure(*args, **kwargs) -> Figure:
...
-    backend = _get_backend_mod()
-    manager = backend.new_figure_manager(1, *args, FigureClass=Figure, **kwargs)
-    return manager.canvas.figure
+    fig = plt.Figure(*args, **kwargs)
+    if is_interactive_backend():
+        # Create a manager for the figure, which makes it interactive, as well as
+        # making it possible to show the figure from the terminal.
+        try:
+            plt._backend_mod.new_figure_manager_given_figure(1, fig)
+        except AttributeError:
```
MridulS, line 62: When does this happen?

src/plopp/backends/matplotlib/utils.py
nvaytet: Thanks I was looking for something like that, although it does not include the `ipympl` backend...

src/plopp/backends/matplotlib/utils.py
MridulS: That's fine I guess :)

nvaytet: This is the sequence I can reproduce, using pure MPL:
```
(base) nvaytet@this:~/code/plopp/jupyter$ ipython
Python 3.11.9 | packaged by conda-forge | (main, Apr 19 2024, 18:36:13) [GCC 12.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.26.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import matplotlib.pyplot as plt

In [2]: fig = plt.Figure()

In [3]: plt._backend_mod.new_figure_manager_given_figure(1, fig)
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 plt._backend_mod.new_figure_manager_given_figure(1, fig)

AttributeError: 'NoneType' object has no attribute 'new_figure_manager_given_figure'

In [4]: plt.show()

In [5]: plt._backend_mod.new_figure_manager_given_figure(1, fig)
Out[5]: <matplotlib.backends.backend_qt.FigureManagerQT at 0x7da165f8cbd0>
```
So until I call `plt.show()`, the `_backend_mod` seems to be `None`?

That said, I now cannot get this error to show up again when using Plopp figures.
Maybe it was related to the backend not being interactive when it should have been (i.e. the fix in `is_interactice_backend` above).
So I'll remove the `try/except` block.

src/plopp/backends/matplotlib/utils.py
nvaytet: Apparently, the `backend_registry` does not exist in older versions of `matplotlib` (at least it's not there in 3.7.3).
So I'm just going to stick with my custom list for now.

nvaytet:
@MridulS any more comments on this one?

MridulS (approved):

PR #385, opened 2024-10-28
Refactor colormapper update mechanism

nvaytet:
When creating e.g. a `scatter3d` plot, the mechanism was to first create a scatter with uniformly colored points, and then override the colors in a second step.
The second step was performed by the colormapper.
This meant that when we had a scatter plot with a large number of points (e.g. the DREAM instrument view), when using the spatial clipping tool, one would first see the points flash with a solid color before changing back to the colors from the colorbar.

This flickering was annoying and confusing.

This refactor changes the logic around. The colormapper no longer sets colors on the artists, the artists know what data they want to display and they ask the colormapper what colors should be used.
I think the logic is overall cleaner.

If we now manually change the `vmin` or `vmax` limits on the colormapper, it will notify the artists that its state has changed, and the artists in turn request new colors.


src/plopp/graphics/colormapper.py
```diff
@@ -177,56 +178,69 @@ def autoscale(self):
...
+        Get or set the maximum value of the colorbar.
         """
-        if self.set_colors_on_update:
-            self._set_artists_colors()
+        return self._vmax
+
+    @vmax.setter
+    def vmax(self, vmax: sc.Variable | float):
+        self._vmax = maybe_variable_to_number(vmax, unit=self._unit)
+        self.apply_limits()
```
jl-wynen, line 243: Given that these properties trigger a computation when they are set, does it make sense to add a method that sets both limits? This could avoid one call to `apply_limits` when both are changed. Basically, I'm asking how often this case happens.

src/plopp/backends/matplotlib/image.py
```diff
@@ -76,14 +81,18 @@ class Image:
...
+        uid: str,
         canvas: Canvas,
+        colormapper: ColorMapper,
         data: sc.DataArray,
         shading: str = 'auto',
         rasterized: bool = True,
         **kwargs,
     ):
         check_ndim(data, ndim=2, origin='Image')
+        self.uid = uid
```
jl-wynen, line 93: Why did you make `uid` a function argument? It seems risky to require the user to provide unique ids.

jl-wynen (commented):
:+1: for the overall idea.

src/plopp/graphics/colormapper.py
nvaytet: For now, I don't think this is actually ever used. When I was implementing, I thought it could be useful to manually set the limits on the colormapper after the figure is created, but we currently have no such examples in practice, the docs or even tests.

Maybe I should just remove the `vmin` and `vmax` properties for now, and only add them when we have a need for them? We will just rely on kwargs from the user to set the limits when the plot is first created.

jl-wynen: I am fine with leaving them or removing them. If you don't see a use case in the (near) future, maybe just remove them.

src/plopp/backends/matplotlib/image.py
jl-wynen: > I has now become useful to have ids so that the colormapper could easily access or remove artists under its management (for example when you remove a set of scatter points).

Would this not be possible with an interface like this? That is, the id is generated by the artist but accessed by the colormapper?
```python
artist = make_artist()
self.artists[artist.uid] = artist
```

> But maybe there is a better way of achieving this? The colormapper could have a list of artists instead of a dict. Maybe we almost never need to remove artists?

Depends on how the artists are usually accessed. If we rarely need to access a specific one by id but usually just loop over all, then a list would be fine.

nvaytet: > Would this not be possible with an interface like this?

This would be ok for the colormapper, but it wouldn't work for the `GraphicalView` as for consistency, we need the artist `uid` to be the same as the `key` passed to the [update function](https://github.com/scipp/plopp/blob/main/src/plopp/graphics/graphicalview.py#L126).
That key is either the name of an entry in a dict/DataGroup, or the `uid` of the `Node` that is providing the data.

How about keeping the `uid` argument, but have a default `None` value. Inside the artist, if it is `None`, we then generate an id from `uuid`?

jl-wynen (approved):

PR #393, opened 2024-11-19
Optimized fast image for linspace coords

### LINKED ISSUE #392: Pcolormesh is slow: add an imshow optimized branch for linspace bins
We are using Matplotlib's `pcolormesh` to plot all images.
Updating colors on the mesh at a high rate is difficult.

We should add an optimized branch in the `Image` class that uses `imshow` if the bins of the data are linspace.

nvaytet:
This adds two branches for plotting images: the old pcolormesh that can handle any coords, and a fast optimized branch that uses imshow to show images with linspace coords.

Fixes #392


src/plopp/backends/matplotlib/mesh_image.py
```diff
@@ -0,0 +1,249 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
nvaytet, line 1: This file was moved basically without changes

src/plopp/backends/matplotlib/image.py
```diff
@@ -1,251 +1,33 @@
...
-        canvas: Canvas,
-        colormapper: ColorMapper,
-        data: sc.DataArray,
-        uid: str | None = None,
-        shading: str = 'auto',
-        rasterized: bool = True,
-        **kwargs,
+    if all(
+        (data.coords[dim].ndim < 2)
+        and ((data.coords[dim].dtype == str) or (sc.islinspace(data.coords[dim])))
```
jl-wynen, line 28: This requires that there are dim-coords for each dim. (Because it uses `data.coords[dim]`.) Was this a requirement already before?

src/plopp/backends/matplotlib/image.py
```diff
@@ -1,251 +1,33 @@
...
-        self,
-        canvas: Canvas,
-        colormapper: ColorMapper,
-        data: sc.DataArray,
-        uid: str | None = None,
-        shading: str = 'auto',
-        rasterized: bool = True,
-        **kwargs,
+    if all(
+        (data.coords[dim].ndim < 2)
```
jl-wynen, line 27: Can `FastImage` handle scalar coords?

src/plopp/backends/matplotlib/image.py
nvaytet: Yes. If the input is missing them, some dummy (arange) coords are added before you get to here.
There are tests for this. See e.g. https://github.com/scipp/plopp/blob/69a60a3a12c699a0489f6f4a084eb07aba7b96ce/tests/plotting/plot_1d_test.py#L51

src/plopp/backends/matplotlib/image.py
nvaytet: If there is a scalar coord that does not have the same name as a dimension
```Py
da = pp.data.data2d()
da.coords['scalar'] = sc.scalar(333., unit='K')
da.plot()
```
is fine (the 'scalar' coord is just ignored.

Interestingly, if the scalar coord is named 'x'
```Py
da = pp.data.data2d()
da.coords['x'] = sc.scalar(333., unit='K')
da
```
it fails, but also fails on `main`.

I can fix in another PR...

jl-wynen (approved):

src/plopp/backends/matplotlib/image.py
nvaytet: See #425

PR #395, opened 2024-11-29
Make pythreejs a soft dependency

nvaytet:
When `%matplotlib widget` is enabled, an interactive figure is made and this imports widgets, and there is a top-level import of `pythreejs` in one of the widgets, thus making `pythreejs` a hard dependency.

We fix this here.

### DIFF
--- a/src/plopp/widgets/clip3d.py
+++ b/src/plopp/widgets/clip3d.py
@@ -8,7 +8,6 @@ from typing import Any, Literal
 
 import ipywidgets as ipw
 import numpy as np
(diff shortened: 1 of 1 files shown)


jl-wynen (approved):

SimonHeybrock:
Will this allow for removing the (I think related) warning filters in our `pyproject.toml`s?

nvaytet:
> Will this allow for removing the (I think related) warning filters in our `pyproject.toml`s?

Possibly?

PR #396, opened 2024-11-29
Expose an extra node of a drawing tool for merging all drawings.

YooSunYoung:
Instead of making a new tool that duplicates a lot, 
I just added a default node that returns all drawings so that you can attach another node that needs all drawings at once.

I added an example how to use the all-drawings-node as a part of rectangle selection tool.


docs/gallery/rectangle-selection.ipynb
```diff
@@ -222,11 +222,116 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "14",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from plopp.widgets import Box\n",
```
nvaytet, line 277: I don't think the changes in the `drawing.py` code below are necessary. It's also adding more callbacks, potentially slowing things down?

I think this was the intended use case for using a node as the `destination` instead of a figure.
After defining the `vertical_sum_all` in your notebook, you can have something like:
```Py
from plopp.widgets import Box

data_node = pp.Node(da)

f2d = pp.imagefigure(data_node, norm='log')

all_rectangles_node = pp.Node(lambda *args: args)
merge_node = pp.Node(vertical_sum_all, da=data_node, rect_infos=all_rectangles_node)

# Trick: Note the func=lambda da, x: x
r = RectangleTool(figure=f2d, input_node=data_node, func=lambda da, x: x, destination=all_rectangles_node)
f2d.toolbar['roi'] = r

f1d = pp.linefigure(merge_node)

box = Box([f2d, f1d])
box
```
I think it achieves the same?

docs/gallery/rectangle-selection.ipynb
```diff
@@ -222,11 +222,116 @@
...
+    "            (da.coords['x'] > max(rect_info['x']['left'], rect_info['x']['right']))\n",
+    "            | (da.coords['x'] < min(rect_info['x']['left'], rect_info['x']['right']))\n",
+    "            | (da.coords['y'] > max(rect_info['y']['bottom'], rect_info['y']['top']))\n",
+    "            | (da.coords['y'] < min(rect_info['y']['bottom'], rect_info['y']['top']))\n",
+    "            for rect_info in rect_infos.values()\n",
+    "        ]\n",
+    "        mask = reduce(lambda x, y: x & y, masks)\n",
+    "        selected_sample_region = da.copy(deep=False)\n",
+    "        selected_sample_region.masks['roi'] = mask['x', :-1]['y', :-1]\n",
+    "        return selected_sample_region.sum(da.dims[-1])\n",
```
nvaytet, line 265:
```suggestion
    "        return selected_sample_region.sum(da.dims[0])\n",
```

docs/gallery/rectangle-selection.ipynb
```diff
@@ -222,11 +222,116 @@
...
+    "    else:\n",
+    "        return da.sum(da.dims[-1])"
```
nvaytet, line 267:
```suggestion
    "        return da.sum(da.dims[0])"
```

docs/gallery/rectangle-selection.ipynb
YooSunYoung: What if we want to see all individual plots and merged plot at the same time?

YooSunYoung: And indeed, maybe the notify children should be handled in a different way...

YooSunYoung: And I think each drawing node will call the `all_rectangles_node` individually so it won't work...?

nvaytet: > What if we want to see all individual plots and merged plot at the same time?

I think that is still possible? You can manually attach a figure to the `all_rectangles_node` (before the `merge_node`)?

YooSunYoung: I don't think I understood. Can you draw a graph for me...??

YooSunYoung:
I think it's not relevant anymore as we can just use the DrawingTool as it is for making merged drawing tool.

PR #398, opened 2024-12-02
Browser based testing environment

YooSunYoung:
I think we can maybe make the test into a tox test... just wanted to make it work in the CI at least.

Didn't write any actual tests yet.


YooSunYoung:
It's moved to beamlime.

PR #400, opened 2024-12-10
Allow lazy update option.

YooSunYoung:


src/plopp/widgets/drawing.py
```diff
@@ -55,6 +63,7 @@ def __init__(
         destination: FigureLike | Node,
         get_artist_info: Callable,
         value: bool = False,
+        lazy_update: bool = False,
```
YooSunYoung, line 66: Can we have this option in the constructor?
(I opened a PR to share the idea.)

nvaytet: Good idea. Should we use the `ipywidgets` terminology (which some users may already know)?; i.e. use `continuous_update = True` instead?

YooSunYoung: Sure. I wasn't aware of the common terminology thanks!

YooSunYoung: @nvaytet Done...!

I just tested in the notebook. Seems working as expected.

https://github.com/user-attachments/assets/b813bcd9-eba4-4138-b113-2ef1fca5490f

nvaytet (approved):

PR #402, opened 2024-12-12
Run conda-build also on main CI

nvaytet:
To try and avoid having conda builds almost always broken during releases, we also build wheels and conda packages during main builds.


MridulS:
We do weekly conda builds for scipp, maybe we can move this to copier_template https://github.com/scipp/scipp/blob/main/.github/workflows/weekly.yml so we are running conda build all across?

PR #403, opened 2025-01-20
fix: add return statements

jokasimr:
Fixes an issue in the `Camera` class where missing return values in properties makes it seem like the properties are `None`.


src/plopp/graphics/camera.py
```diff
@@ -122,31 +122,31 @@ def position(self) -> tuple[float, float, float] | None:
         The position of the camera. If camera units have been set, the position returned
         will be converted to the specified units.
         """
-        self.get('position')
+        return self.get('position')
 
     @property
     def look_at(self) -> tuple[float, float, float] | None:
```
nvaytet, line 128: Interesting. I wonder if the `| None` in the return types can be removed? Maybe they were added because `mypy` realised we were not returning anything?

jokasimr: `self.get` also returns `| None` and internally `self.get` uses `dict.get` that also returns `| None` so it's not possible to remove `|None` here in a way that keeps mypy happy.

Do we know that `self.get('position')` will always return something non-`None`? I'm not familiar enough with the code to tell

nvaytet: Ok, it's probably fine as it is.

nvaytet (approved):

PR #410, opened 2025-02-12
Copier and tox deps bump

MridulS:


MridulS:
The failures here are from deprecation warnings from plotly.

jl-wynen:
:+1: for the change as such. But the warnings should be fixed before merging this.

pyproject.toml
```diff
@@ -84,6 +84,7 @@ filterwarnings = [
   'ignore:setDaemon\(\) is deprecated, set the daemon attribute instead:DeprecationWarning',
   'ignore:TriCutTool is deprecated::',
   'ignore:Cut3dTool is deprecated::',
+  'ignore: *scattermapbox* is deprecated:DeprecationWarning',
```
jl-wynen, line 87: Not what I meant. We should actually change the implementation to work with the new API. Otherwise it will just break without warning.

jl-wynen (changes requested):

jl-wynen (commented):
Never mind, should have read your message in the issue first. But please add a comment in pyproject.toml linking to the upstream issue.

MridulS:
The text for expected test needs to be updated `Tests (ubuntu-22.04, 3.10, py310) / test` to `Tests (ubuntu-24.04, 3.10, py310) / test`

jl-wynen (approved):

PR #418, opened 2025-03-05
Fix autoscale behaviour for slicer plots (including 3d)

nvaytet:
Fixes https://github.com/scipp/essdiffraction/issues/134

### DIFF
--- a/src/plopp/graphics/graphicalview.py
+++ b/src/plopp/graphics/graphicalview.py
@@ -52,6 +52,7 @@ class GraphicalView(View):
         format: Literal['svg', 'png'] | None = None,
         legend: bool | tuple[float, float] = False,
         camera: Camera | None = None,
+        autoscale: bool = True,
         ax: Any = None,
         cax: Any = None,
         **kwargs,
@@ -67,6 +68,7 @@ class GraphicalView(View):
         self.draw_on_update = True
         self._data_name = None
         self._data_axis = None
+        self._autoscale = autoscale
 
         self.canvas = canvas_maker(
             cbar=cbar,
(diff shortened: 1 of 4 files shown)


src/plopp/graphics/graphicalview.py
```diff
@@ -208,15 +214,20 @@ def update(self, *args, **kwargs) -> None:
         if need_legend_update:
             self.canvas.update_legend()
 
-        if self.draw_on_update:
```
jl-wynen, line 211: Looks like `draw_on_update` can be removed.

src/plopp/graphics/graphicalview.py
```diff
@@ -229,10 +240,11 @@ def render(self) -> None:
         all parent nodes and draw the figure.
         In addition, we call the home method to autoscale the axes and colormapper.
         """
-        self.draw_on_update = False
```
jl-wynen, line 232: Does this now trigger two redraws because drawing is no longer disabled in `update`?

nvaytet: I think it does only one because of the `_autoscale_or_draw`?

jl-wynen: I'm not familiar enough with what happens in `super().render()`. But it looks to me like it would call `self.update`. And that calls `_autoscale_or_draw` which, because `self._autoscale == False`, draws. And then it calls `self.fit_to_data()` which draws a second time. Which one did I get wrong?

nvaytet: I think you're right.

I changed the logic slightly, but to keep things simple, I'm still making multiple draw calls when `render` is called.
This seems to me as overkill to try and prevent that, as it's only called once when the figure is created, and only if there are multiple lines/artists.
It is not called at a high rate, so I think it's fine.

I can change it to only make a single call if you like, but the code will be more confusing I think.

jl-wynen:
This looks good to me now. But I don't understand how the failing tests are related to this change.

jl-wynen (approved):

PR #429, opened 2025-03-27
Replace the hacky `_plopp_observe` by using `ValueWidget`

### LINKED ISSUE #406: Use composite widgets instead of only inheriting from VBox/HBox
Our custom widgets have a `_plopp_observe` method instead of `observe` because inheriting from `VBox` causes errors when overriding the `observe` method (see e.g. https://github.com/scipp/plopp/blob/main/src/plopp/widgets/slice.py#L73)

However, it is possible to work around this by also inheriting from `ValueWidget`. See https://github.com/jupyter-widgets/ipywidgets/issues/2839#issuecomment-611034761

This would make our implementation cleaner.

nvaytet:
Fixes #406


YooSunYoung (approved):
Cool...!

PR #432, opened 2025-04-03
Change colors of theme and logo in docs

nvaytet:


docs/_static/custom.css
```diff
@@ -0,0 +1,9 @@
+html[data-theme="light"] {
+    --pst-color-primary: #800080;
+    --pst-color-secondary: #ed4c5c;
```
jl-wynen, line 3: :+1: for adjusting the theme colour.

But why did you pick a secondary colour that is so similar to Scitacean?

docs/_static/logo-large-dark.svg
jl-wynen, line 1: A bit sad that the dark mode logo is white. How hard would it be to make a version with an adjusted colour? Is it enough to automatically increase lightness and decrease saturation?

jl-wynen (commented):
No hard blockers here. But should we ask around if everyone is happy with the new colours? Or if someone wants a different colour for their favourite package?

docs/_static/custom.css
nvaytet: > But why did you pick a secondary colour that is so similar to Scitacean?

No idea, I looked at some palettes online and tried to find a colour that would go well.

docs/_static/logo-large-dark.svg
nvaytet: > Is it enough to automatically increase lightness and decrease saturation?

Would be very useful if we could do that for logo and also the dark version of the primary/secondary colors?

nvaytet: > automatically increase lightness and decrease saturation

Can you do that dynamically? (i.e. without making a deparate 'dark' file)

jl-wynen: Sort of. You can apply CSS rules to SVGs. But then you should probably modify the SVG's code to add some classes or ids so that you can identify the elements you want to style.

docs/_static/custom.css
nvaytet: Could pick some blue/green/turquoise color maybe?

nvaytet:
@jl-wynen see update with green color as secondary, and coloured dark logo.

PR #438, opened 2025-04-14
Docs pages reorganisation

nvaytet:
We move things around in the docs to make them more user-friendly.

1. I removed the `user-guide` section, and moved the subfolders at the top level.
1. The landing page looks like the old user-guide index, with the cards and images
1. The Navigation bar now has sections: `Getting started`, `Plotting`, `Custom figures`, `Gallery`
1. Api reference. development, About are now in the drop-down menu
1. Added an icon link to the Scipp organisation (this should be a landing page for the Scipp ecosystem, once we have one)


MridulS:
https://remote-unzip.deno.dev/scipp/plopp/artifacts/2941558734/getting-started/overview.html there is an error `MemoryError: std::bad_alloc`  on the rendered doc page.

nvaytet:
Nice catch!
My guess is that this is the PR that broke it... #393

nvaytet:
> Help needed: how to turn the Scipp icon in the top right to being the same color as the other icons?

@jl-wynen can you help me with this please? :pray:

PR #441, opened 2025-04-23
Fix zorder in 1d plots, and masks on non bin-edge lines when solid linestyle is requested

### LINKED ISSUE #440: zorder argument fails in line plot
```Py
import plopp as pp

da = pp.data.data1d()
da.plot(zorder=10)
```
fails with
```
File [~/code/path/plopp/backends/matplotlib/line.py:112](http://localhost:8889/home/nvaytet/code/path/plopp/backends/matplotlib/line.py#line=111), in Line.__init__(self, canvas, data, uid, artist_number, errorbars, mask_color, **kwargs)
    110     self._mask.set_visible(line_data['mask']['visible'])
    111 else:
--> 112     self._line = self._ax.plot(
    113         line_data['values']['x'],
    114         line_data['values']['y'],
    115         label=self.label,
    116         zorder=10,
    117         **{**default_plot_style, **line_args},
    118     )[0]
    119     self._mask = self._ax.plot(
    120         line_data['mask']['x'],
    121         line_data['mask']['y'],
   (...)
    128         visible=line_data['mask']['visible'],
    129     )[0]
    131 self.line_mask = sc.array(dims=['x'], values=~np.isnan(line_data['mask']['y']))

TypeError: matplotlib.axes._axes.Axes.plot() got multiple values for keyword argument 'zorder'
```

nvaytet:
Fixes #440

### DIFF
--- a/src/plopp/backends/matplotlib/line.py
+++ b/src/plopp/backends/matplotlib/line.py
@@ -103,12 +103,6 @@ class Line:
             self._mask = self._ax.step(line_data['mask']['x'], line_data['mask']['y'])[
                 0
             ]
-            self._mask.update_from(self._line)
-            self._mask.set_color(mask_color)
-            self._mask.set_label(None)
-            self._mask.set_linewidth(self._mask.get_linewidth() * 3)
-            self._mask.set_zorder(self._mask.get_zorder() - 1)
-            self._mask.set_visible(line_data['mask']['visible'])
         else:
             self._line = self._ax.plot(
                 line_data['values']['x'],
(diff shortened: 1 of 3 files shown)


nvaytet:
@g5t does this fix your issue?

src/plopp/backends/matplotlib/line.py
```diff
@@ -153,7 +150,6 @@ def update(self, new_values: sc.DataArray):
         check_ndim(new_values, ndim=1, origin='Line')
         self._data = new_values
         line_data = make_line_data(data=self._data, dim=self._dim)
-        self.line_mask = sc.array(dims=['x'], values=~np.isnan(line_data['mask']['y']))
```
nvaytet, line 156: This seems to be left over from before, and not used anywhere...

g5t:
I didn't look at the resulting figures closely enough :facepalm: 
While `zorder` can be set in the call to `plot`, its value is not respected and when set on the 'parasitic' axes directly the 'main' axes `facecolor` must still be set to `'None'`.

The attached notebook code runs without errors, but only the last cell produces the expected output.
[plopp_zorder.ipynb.zip](https://github.com/user-attachments/files/19871995/plopp_zorder.ipynb.zip)

nvaytet:
Thanks for the update.
As far as I can see, this appears to be an issue when using `twinx`.
You have the same issues with pure matplotlib:
```Py

fig, ax1 = plt.subplots()
ax2 = ax1.twinx()

fm = ax1.plot(x, y1.values, linestyle='solid', color='red', marker=None, linewidth=5, zorder=10)
fp = ax2.plot(x, y2.values, linestyle='solid', color='black', marker=None, linewidth=5)
```
Screenshot_20250424_104026

An alternative to your last solution could be to plot the black curve first, and swap the axes labels left and right, thsu removing the need to mess with the axes facecolor:
```Py
fp = y2.plot(linestyle='solid', color='black', marker=None, linewidth=5, vmin=-3, vmax=3)
fm = y1.plot(ax=fp.ax.twinx(), linestyle='solid', color='red', marker=None, linewidth=5)
fp.ax.yaxis.tick_right()
fp.ax.yaxis.set_label_position("right")
fm.ax.yaxis.tick_left()
fm.ax.yaxis.set_label_position("left")

fm.ax.set(title='Right z-order w/o face color', ylabel=r'$\sin(20 \pi x)$')
fm.ax.yaxis.label.set(color='red')
fm.ax.yaxis.set_tick_params(color='red', labelcolor='red')
fp.ax.set(ylabel=r'$\cos(2 \pi x)$')
fm
```
But it's even more lines of code, so not sure it's worth it.

I think the point is that you find a solution that works, and it appears to me a MPL-specific issue, so there isn't much more we can do here in Plopp...

PR #442, opened 2025-04-23
Update deprecation warnings filter in pyproject

### LINKED ISSUE #421: DeprecationWarning from asyncio in debounce mechanism for 3d clip planes
The warning raised is
```
../../path/plopp/widgets/debounce.py:23: in start
    self._task = asyncio.ensure_future(self._job())
../../../miniforge3/lib/python3.12/asyncio/tasks.py:693: in ensure_future
    loop = events.get_event_loop()

DeprecationWarning: There is no current event loop
```

This happens when running the unit tests on python 3.12.

I do not see the warning (even in the jupyter log console) when using the clipping planes in the notebook.
It could be a difference coming from the fact that the widget are never displayed when running the unit tests?

This makes a difference in some cases, for example the [play widget](https://ipywidgets.readthedocs.io/en/8.0.5/examples/Widget%20List.html#play-animation-widget) does not function properly if it is not displayed in a notebook.

nvaytet:
- Fixes #421 
- Also remove some old filters which are from classes that have now been removed.

This is not technically a fix for #421 but at least tests will continue to pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,10 +81,9 @@ filterwarnings = [
   'ignore:Passing unrecognized arguments to super:DeprecationWarning',
   'ignore:Jupyter is migrating its paths:DeprecationWarning',
   'ignore:setDaemon\(\) is deprecated, set the daemon attribute instead:DeprecationWarning',
-  'ignore:TriCutTool is deprecated::',
-  'ignore:Cut3dTool is deprecated::',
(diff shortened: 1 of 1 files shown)


jl-wynen (approved):
Ok. But if, according to #421, there is a difference between displaying widgets and not, are you sure that the tests actually test the relevant code?

PR #448, opened 2025-05-14
Drop unused coords

jl-wynen:
AFAIK, there is no need to keep all coords around. And this caused a problem in Scipp where the averaging introduced in #428 fails for a datetime coord. (See https://github.com/scipp/scipp/pull/3695)

### DIFF
--- a/src/plopp/plotting/common.py
+++ b/src/plopp/plotting/common.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
 import warnings
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from typing import Any
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


src/plopp/plotting/common.py
```diff
@@ -192,7 +214,12 @@ def preprocess(
     coords:
         If supplied, use these coords instead of the input's dimension coordinates.
     """
-    out = to_data_array(obj)
+    if isinstance(coords, str):
+        coords = [coords]
```
nvaytet, line 218: I think this means that the check
```Py
        if isinstance(coords, str):
            coords = [coords]
```
further down (L231) can be removed?

tests/plotting/common_test.py
```diff
@@ -70,3 +70,19 @@ def test_preprocess_raises_for_unsupported_dtype(dtype_and_shape):
...
+    da.coords['yy2'] = 3.3 * da.coords['yy']
+    out = preprocess(da)
+    assert set(out.coords) == {'xx', 'yy'}
+
+
+def test_preprocess_drops_coords_that_are_not_plotted_custom_coords():
+    da = data_array(ndim=2)
+    da.coords['xx2'] = 7.5 * da.coords['xx']
+    da.coords['yy2'] = 3.3 * da.coords['yy']
+    out = preprocess(da, coords=['yy2', 'xx'])
```
nvaytet, line 87: Also test the case where the arg passed as `coords` is a string?
e.g. `preprocess(da, coords='yy2')`

src/plopp/plotting/common.py
jl-wynen: Yes, removed

tests/plotting/common_test.py
jl-wynen: Done

nvaytet (approved):

PR #467, opened 2025-07-07
Add player to slicing slider widgets

### LINKED ISSUE #462: Play button for slicer
It was requested from many instrument teams in the live data reduction meetings. (also for presentation materials from IDS members)

What they asked is to be able to have an animation of past recordings but going back to past of kafka messages is not necessary since we have files.

What they want to see is (mostly) how the recorded data changes over wall-clock-time or over time-of-flight.
A play button from a slider should be enough.

nvaytet:
Example:
```Py
import plopp as pp

da = pp.data.data3d()
pp.slicer(da, keep=['x', 'y'], enable_player=True)
```
Screenshot_20250707_161139

Fixes #462 

Note that I decided to disable it for `RangeSlider` because it would really complicate things, especially when the looping is enabled, then the max would temporarily become the min, etc...

### DIFF
--- a/docs/plotting/slicer-plot.ipynb
+++ b/docs/plotting/slicer-plot.ipynb
@@ -64,7 +64,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "fba6798e-6843-4b82-b304-822e2277dfe8",
+   "id": "6",
    "metadata": {},
    "source": [
     "## Animate the sliders\n",
@@ -84,7 +84,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "31a1e511-951e-4d91-a5e1-ef7f8afbb59d",
+   "id": "7",
    "metadata": {},
    "outputs": [],
    "source": [
--- a/docs/plotting/slicer-plot.ipynb
+++ b/docs/plotting/slicer-plot.ipynb
@@ -90,6 +90,19 @@
    "source": [
     "pp.slicer(da, keep=['x', 'y'], enable_player=True)"
    ]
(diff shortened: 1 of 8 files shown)


jl-wynen (commented):
Looks good, but two requests about the docs:

- Can you show the new argument in the slicer docs page?
- And can you link from the scatter 3d gallery to the slicer docs? I first went to the gallery to test the feature but couldn't find any examples.

jl-wynen:
> > * And can you link from the scatter 3d gallery to the slicer docs? I first went to the gallery to test the feature but couldn't find any examples.
> 
> Can you clarify what you wanted? Did you want a link from the [Scatter3d plot with slider](https://scipp.github.io/plopp/gallery/scatter3d-with-slider.html) to the docs page for the [slicer plot](https://scipp.github.io/plopp/plotting/slicer-plot.html)? In which case I'm not sure it's super related as the slicer plot is only for 1d/2d data. The example in the gallery is basically the answer to the question "can I also make this plot with a 3d scatter?"
> 
> Ah, maybe you meant "can you link from the slicer plot docs to the gallery example" to answer this exact question?
> 
> Or maybe you meant "can i link to the API reference of the `SlicerWidget` from the scatter3d with slicer gallery example?

I'm not sure myself anymore. I think I wanted to link from [Scatter3d plot with slider](https://scipp.github.io/plopp/gallery/scatter3d-with-slider.html). But looking at it now, I agree that it doesn't make sense.

I think it is fine the way it is.

jl-wynen (approved):

PR #471, opened 2025-07-15
deps: lower pins and tox -e deps

### LINKED ISSUE #463: Add minimum Scipp requirement?
According to `blame` this was introduced 2 weeks ago. I think it requires https://github.com/scipp/scipp/releases/tag/25.04.0 now as a minimum? Or am I missing something?

```python
TypeError                                 Traceback (most recent call last)
Cell In[5], line 6
      4 # Load only first run
      5 first_run = sans_workflow.compute(loaded[0])
----> 6 sc.DataGroup(sc.collapse(first_run['data'], keep='tof')).plot()

File ~/.conda/envs/mantid_env/lib/python3.10/site-packages/scipp/core/data_group.py:549, in DataGroup.plot(self, *args, **kwargs)
    546 def plot(self, *args: Any, **kwargs: Any) -> Any:
    547     import plopp
--> 549     return plopp.plot(self, *args, **kwargs)

File ~/.conda/envs/mantid_env/lib/python3.10/site-packages/plopp/plotting/plot.py:105, in plot(obj, aspect, cbar, coords, errorbars, figsize, grid, ignore_size, mask_color, norm, scale, title, vmin, vmax, legend, **kwargs)
    103 ndims = set()
    104 for n in nodes:
--> 105     ndims.add(n().ndim)
    106 if len(ndims) > 1:
    107     raise ValueError(
    108         'All items given to the plot function must have the same '
    109         f'number of dimensions. Found dimensions {ndims}.'
    110     )

File ~/.conda/envs/mantid_env/lib/python3.10/site-packages/plopp/core/node_class.py:77, in Node.__call__(self)
     76 def __call__(self) -> Any:
---> 77     return self.request_data()

File ~/.conda/envs/mantid_env/lib/python3.10/site-packages/plopp/core/node_class.py:137, in Node.request_data(self)
    133     args = (parent.request_data() for parent in self.parents)
    134     kwargs = {
    135         key: parent.request_data() for key, parent in self.kwparents.items()
    136     }
--> 137     self._data = self.func(*args, **kwargs)
    138 return self._data

File ~/.conda/envs/mantid_env/lib/python3.10/site-packages/plopp/plotting/common.py:295, in preprocess(obj, name, ignore_size, coords)
    293 out = _add_missing_dimension_coords(out)
    294 out = _drop_non_dimension_coords(out)
--> 295 out = _handle_coords_with_left_over_dimensions(out)
    296 _check_coord_sanity(out)
    297 return out

File ~/.conda/envs/mantid_env/lib/python3.10/site-packages/plopp/plotting/common.py:214, in _handle_coords_with_left_over_dimensions(da)
    202 def _handle_coords_with_left_over_dimensions(da: sc.DataArray) -> sc.DataArray:
    203     """
    204     In some rare cases, the coordinate may have dimensions that are not present in the
    205     data array. This can happen for example when a 2d coord with bin edges is sliced
   (...)
    211     dimension that is not present in the data array.
    212     """
    213     return da.assign_coords(
--> 214         {
    215             name: da.coords[name].mean(
    216                 [dim for dim in da.coords[name].dims if dim not in da.dims]
    217             )
    218             for name in da.coords
    219         }
    220     )

File ~/.conda/envs/mantid_env/lib/python3.10/site-packages/plopp/plotting/common.py:215, in <dictcomp>(.0)
    202 def _handle_coords_with_left_over_dimensions(da: sc.DataArray) -> sc.DataArray:
    203     """
    204     In some rare cases, the coordinate may have dimensions that are not present in the
    205     data array. This can happen for example when a 2d coord with bin edges is sliced
   (...)
    211     dimension that is not present in the data array.
    212     """
    213     return da.assign_coords(
    214         {
--> 215             name: da.coords[name].mean(
    216                 [dim for dim in da.coords[name].dims if dim not in da.dims]
    217             )
    218             for name in da.coords
    219         }
    220     )

File ~/.conda/envs/mantid_env/lib/python3.10/site-packages/scipp/core/reduction.py:67, in mean(x, dim)
     65     return _call_cpp_func(_cpp.mean, x)  # type: ignore[return-value]
     66 else:
---> 67     return _call_cpp_func(_cpp.mean, x, dim=dim)

File ~/.conda/envs/mantid_env/lib/python3.10/site-packages/scipp/core/_cpp_wrapper_util.py:27, in call_func(func, *args, **kwargs)
     25     return data_group_nary(func, *args, **kwargs)
     26 if out is None:
---> 27     return func(*args, **kwargs)
     28 else:
     29     return func(*args, **kwargs, out=out)

TypeError: mean(): incompatible function arguments. The following argument types are supported:
    1. (x: scipp._scipp.core.Variable) -> scipp._scipp.core.Variable
    2. (x: scipp._scipp.core.Variable, dim: str) -> scipp._scipp.core.Variable
    3. (x: scipp._scipp.core.DataArray) -> scipp._scipp.core.DataArray
    4. (x: scipp._scipp.core.DataArray, dim: str) -> scipp._scipp.core.DataArray
    5. (x: scipp._scipp.core.Dataset) -> scipp._scipp.core.Dataset
    6. (x: scipp._scipp.core.Dataset, dim: str) -> scipp._scipp.core.Dataset

Invoked with: <scipp.Variable> (tof: 501)    float64            [µs]  [5, 205, ..., 99805, 100000]; kwargs: dim=[]
```

jokasimr:
Fixes #463


pyproject.toml
```diff
@@ -31,31 +31,31 @@ requires-python = ">=3.10"
...
   "matplotlib>=3.5",
 ]
 
 dynamic = ["version"]
 
 [project.optional-dependencies]
-scipp = ["scipp"]
-all = ["scipp", "ipympl", "pythreejs", "mpltoolbox", "ipywidgets", "graphviz"]
+scipp = ["scipp>=25.5.0"]
+all = ["scipp>=25.5.0", "ipympl>0.9.4", "pythreejs>=2.4.1", "mpltoolbox>=24.6.0", "ipywidgets>=8.1.7", "graphviz>=0.20.3"]
```
nvaytet, line 42: I'd like to push back a little on this approach...
While I think it's a good idea to have lower bounds for the packages that we maintain ourselves, adding lower bounds everywhere feels like it could bring unforseen problems.

Or at least, we should go back and find the lowest version of every dependency that is compatible with our package, instead of just adding the version number that was solved by the last run of `tox -e deps`.

Say there is another package, not in our dependencies, but required by the conda env on a VISA instance (e.g. an analysis package) that has added an upper bound (from some reason) on e.g. `ipympl<0.9.3` (or maybe `numpy<2` would be a common one because packages may not yet have migrated).
So now, we have an environment which is unsolvable, while actually for us, having `ipympl==0.9.1` would have been completely fine.

Generally, the less bounds we have the better?
Unless I am missing something?

jl-wynen: > I see it as a fine balance. If you can't fx import plopp because it fails because of a dependency, then we should absolutely have a lower bound. Same if you just can't plot anything.
If there are just a couple of glitches, or not all the features are fully working, then we should avoid the lower bound?

I disagree about the last point here. If a package can be installed, it should also fully work and not crash at runtime because of some dependency conflict. (Modulo optional features that need to be handled with a meaningful exception.) If this was not the case, users could not trust that our packages work. And we would also not get any indication that there is a problem until users come to us with cryptic error messages.

If we want to be backwards compatible, we should handle both new and old interfaces of a library explicitly. In general, we should be careful with using new features early.

And if a user fails to resolve an env but they still want to use the conflicting packages together, they can override transitive dependencies. (At least with uv. Not sure whether pixi supports it yet. Pip and conda don't, to my knowledge.)

jokasimr:
I fiddled a bit with the pins and managed to relax some of them.
The matplotlib pin had to be increased to make the tests pass, I didn't notice that before because I didn't remove the `uv.lock` file in between test runs.

pyproject.toml
jokasimr: > otherwise some of them resolve to something ridiculous like 0.0.1-alpha-use-at-your-own-peril.
> is something that happens in practice, then I agree with the 1-2 year old approach

If you ask uv to find the earliest version of dependencies without setting any constraints on them that is what will happen.

If some package in VISA pins, for example, `ipywidgets<8` and the tests in plopp don't pass with `ipywidget<8` isn't that something we want to be aware of? That's the purpose of the nightly run. That doesn't mean we necessarily have to update the pins to make the nightly pass, that can be done on a case by case basis. If the dependency is central to the package then it probably makes sense to update the pin, but sometimes it could also make sense to update our package to continue to support the older version of the dependency.

MridulS (approved):

PR #487, opened 2025-09-08
Fix image plot to also accept position updates

### LINKED ISSUE #485: Image plot: coordinates/positions do not update
Doing
```Py
import plopp as pp

%matplotlib widget

n = pp.Node(pp.data.data2d())
f = pp.plot(n)
f
```
and then
```Py
new = pp.data.data2d() * 2.0
new.coords['x'] *= 0.5
n.func = lambda: new
n.notify_children("")
```
updates the colors but not the positions of the points.

Note that for pcolormesh, if the number of points have changed, we probably need to remove the mesh and create a bew one, as in #484

nvaytet:
Fixes #485


src/plopp/backends/matplotlib/image.py
```diff
@@ -1,36 +1,374 @@
...
+                self._bin_edge_coords[k]
+            )
+            if self._data.coords[self._data.dims[i]].dtype == str:
+                self._string_labels[k] = self._data.coords[self._data.dims[i]]
+
+        if self._optimized_mode:
+            self._xmin, self._xmax = self._bin_edge_coords["x"].values[[0, -1]]
+            self._ymin, self._ymax = self._bin_edge_coords["y"].values[[0, -1]]
+            self._dx = np.diff(self._bin_edge_coords["x"].values[:2])
+            self._dy = np.diff(self._bin_edge_coords["y"].values[:2])
```
jl-wynen, line 157: Ideally, you wouldn't define attributes of `self` outside of `__init__`. And I don't like that these attrs only exist for the optimized mode.
Can you use composition to have a `FastImage` or `MeshImage` object inside `Image`? This might also get rid of the `if self._optimized_mode:` checks scattered everywhere.

PR #491, opened 2025-09-30
Simply require a function for a drawing tool

nvaytet:
Instead of requiring a function that returns a function in the `DrawingTool`, we now simply require a plain function.

This means in the rectangle selection example from the gallery, instead of 
```Py
def _get_rect_info(artist, figure):
    return lambda: {
        'x': {
            'dim': figure.canvas.dims['x'],
            'left': sc.scalar(artist.xy[0], unit=figure.canvas.units['x']),
            'right': sc.scalar(
                artist.xy[0] + artist.width, unit=figure.canvas.units['x']
            ),
        },
        'y': {
            'dim': figure.canvas.dims['y'],
            'bottom': sc.scalar(artist.xy[1], unit=figure.canvas.units['y']),
            'top': sc.scalar(
                artist.xy[1] + artist.height, unit=figure.canvas.units['y']
            ),
        },
    }
```
we now have (the `lambda` is gone):
```Py
def _get_rect_info(artist, figure):
    """
    Convert the raw rectangle info to a dict containing the dimensions of
    each axis, and values with units.
    """
    return {
        'x': {
            'dim': figure.canvas.dims['x'],
            'left': sc.scalar(artist.xy[0], unit=figure.canvas.units['x']),
            'right': sc.scalar(
                artist.xy[0] + artist.width, unit=figure.canvas.units['x']
            ),
        },
        'y': {
            'dim': figure.canvas.dims['y'],
            'bottom': sc.scalar(artist.xy[1], unit=figure.canvas.units['y']),
            'top': sc.scalar(
                artist.xy[1] + artist.height, unit=figure.canvas.units['y']
            ),
        },
    }
```

### DIFF
--- a/src/plopp/widgets/drawing.py
+++ b/src/plopp/widgets/drawing.py
@@ -76,6 +76,7 @@ class DrawingTool(ToggleTool):
         self._destination = destination
         self._destination_is_fig = is_figure(self._destination)
         self._get_artist_info = get_artist_info
+        self._get_artist_info_is_callable = None
         self._tool.on_create(self.make_node)
         self._tool.on_remove(self.remove_node)
         if continuous_update:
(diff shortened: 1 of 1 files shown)


src/plopp/widgets/drawing.py
```diff
@@ -105,7 +105,7 @@ def make_node(self, artist):
 
     def update_node(self, artist):
         n = self._draw_nodes[artist.nodeid]
-        n.func = self._get_artist_info(artist=artist, figure=self._figure)
+        n.func = partial(self._get_artist_info, artist=artist, figure=self._figure)
```
nvaytet, line 108: This will break people's notebooks...
Any idea on how to make this retro-compatible?

jokasimr: Is `_get_artist_info` intended to be a function without side effects? If we can rely on that, we could call it and check if it returns a function, if it returns a function we use that instead of applying partial to it.

PR #493, opened 2025-10-03
Update and standardize all top-level function signatures and docstrings

nvaytet:
Alternative to #492.
We just update everything manually here.

It's more difficult to maintain, but means that we can still use Mypy and is more customizable.


tests/graphics/artists_test.py
```diff
@@ -10,68 +10,88 @@
 from plopp.data import examples
 
 CASES = {
-    "line-mpl-static": (('2d', 'mpl-static'), pp.plot, (pp.data.data1d(),)),
-    "line-mpl-interactive": (('2d', 'mpl-interactive'), pp.plot, (pp.data.data1d(),)),
+    "line-mpl-static": (('2d', 'mpl-static'), pp.plot, {'obj': pp.data.data1d()}),
```
nvaytet, line 13: Changes in this file are because `mesh3d` no longer accepts positional arguments

tests/plotting/scatter_test.py
```diff
@@ -112,11 +117,11 @@ def test_scatter_does_not_accept_data_with_other_dimensionality_on_update():
     with pytest.raises(
         sc.DimensionError, match='Scatter only accepts data with 1 dimension'
     ):
-        fig.update(new=data_array(ndim=2, dim_list="xyzab"))
+        fig.update(new=da.fold(dim=da.dim, sizes={'a': 2, 'b': -1}))
```
nvaytet, line 120: It seems that this test was sort of broken. It was getting data with an incompatible unit, as well as incompatible dims.

src/plopp/plotting/slicer.py
```diff
@@ -98,32 +103,60 @@ def __init__(
...
+        self.figure = make_figure(*self.slice_nodes)
         require_interactive_figure(self.figure, 'slicer')
         self.figure.bottom_bar.add(self.slider)
 
 
 def slicer(
     obj: PlottableMulti,
-    *,
     keep: list[str] | None = None,
+    *,
```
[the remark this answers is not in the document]
nvaytet: making it a positional argument was the intention

jl-wynen (approved):

PR #497, opened 2025-11-03
Fix logc formatter error

SimonHeybrock:
I don't know if there is a simpler way of reproducing this outside a notebook?

@nvaytet You probably have a better way to fix this, please feel free to ignore or push to this branch instead!


nvaytet:
I don't think we should try to fix this, I suspect it's a matplotlib issue...

nvaytet:
I believe this was fixed in #500 ? @jokasimr

PR #518, opened 2026-01-29
feat: polygon inspector tool

jokasimr:
There's been requests from time to time for a plopp tool that visualizes intensity inside a polygon region, similar to the `inspector` tool, but with a polygon instead of a point.

Be aware that it is mostly AI-written.

### DIFF
--- a/src/plopp/plotting/inspector.py
+++ b/src/plopp/plotting/inspector.py
@@ -262,7 +262,7 @@ def inspector(
         tool = PolygonTool(
             figure=f2d,
             input_node=bin_edges_node,
-            func=_slice_inside_polygon,
+            func=_mask_outside_polygon,
             destination=f1d,
             tooltip="Activate polygon inspector tool",
         )


src/plopp/widgets/drawing.py
```diff
@@ -153,6 +286,32 @@ def _get_points_info(artist, figure):
     }
 
 
+def _get_polygon_info(verts, figure):
+    """
+    Convert the raw polygon vertices to a dict containing the dimensions of
+    each axis, and arrays with units.
+    """
+    xs, ys = zip(*verts, strict=False)
```
YooSunYoung: [minor] Should this be strict?

jokasimr:
Example usage:

Screenshot from 2026-01-29 12-39-13

src/plopp/plotting/inspector.py
```diff
@@ -34,6 +35,72 @@ def _slice_xy(da: sc.DataArray, xy: dict[str, dict[str, int]]) -> sc.DataArray:
...
+
+    xdim = poly['x']['dim']
+    ydim = poly['y']['dim']
+    x = _coord_to_centers(da, xdim)
+    y = _coord_to_centers(da, ydim)
+    vx = poly['x']['value'].to(unit=x.unit).values
+    vy = poly['y']['value'].to(unit=y.unit).values
+    verts = np.column_stack([vx, vy])
+    path = Path(verts)
+    xx, yy = np.meshgrid(x.values, y.values, indexing='xy')
```
nvaytet: I think this assumes that the x and y coords are 1d arrays, and would break down if one of the coords is 2d.
Instead, I think we can broacast the coord using scipp and then take the values of that?
e.g.
```Py
xx = sc.broadcast(x, sizes=da.sizes).values
yy = sc.broadcast(y, sizes=da.sizes).values
```
and this would be a no-op if the coord is already 2d?

src/plopp/plotting/inspector.py
```diff
@@ -34,6 +35,72 @@ def _slice_xy(da: sc.DataArray, xy: dict[str, dict[str, int]]) -> sc.DataArray:
...
+        # Order of dims after slicing: points, *all dims not sliced by the mask.
+        dims=['points', *(d for d in v.dims if d not in mask.dims)],
+        values=v.values[mask.values].reshape(
+            -1, *(s for d, s in v.sizes.items() if d not in mask.dims)
+        ),
+        unit=v.unit,
+    )
+
+
+def _slice_dataarray_by_mask(da, mask):
```
nvaytet: I remember we discussed whether slicing would be a faster operation than adding a mask, but looking at this I feel the added complexity of having to loop through data, coords and masks might be making things less efficient that simply adding a 2d mask in the x and y dimensions?

We already get the mask computed by the
```Py
inside = sc.array(
        dims=[ydim, xdim], values=path.contains_points(points).reshape(yy.shape)
    )
```
below.
Can we just set the inverse of that as a mask and then do the reduction operation in the x and y dimensions?
I think that should give the same result?

jokasimr: Yeah I think you're right.  Let's do that instead.

PR #528, opened 2026-02-13
Updates to inspector plot: add rectangle mode and fix operation inside polygons

nvaytet:
Some updates to the inspector plot:

- fixed slicing when used on data with a 2d coordinate
- small optimization to avoid converting coords to midpoints every time a polygon is modified
- added rectangle analog to #518 
- use better data in the inspector notebook (clusters3d and air temperature)

Screenshot_20260218_100505

A question arose while implementing this: there is an `operation` argument that decides which reduction operation to apply along the 3rd dimension into the paper. This is a `sum` by default, but can be a `min`, `max` or `mean`.
Now, when we select a region inside a rectangle or a polygon, we need to also reduce that along the x and y dimensions before sending to the 1d plot along z. The operation there is currently always a sum, and I don't think it would make sense to do anything else than sum?

So the question is: does it even make sense to have other operations along the z dimension?

~**Update:** I decided to remove the operation in the end.~

**Edit:** there is an example where performing a `mean` operation actually makes sense...
https://scipp.github.io/plopp/getting-started/numpy-pandas-xarray.html#Interactive-tools

So instead of removing the `operation`, we apply the same operation along the x/y dims inside a selected region.
I think this makes sense: in the 1d plot, we would want to see what is the mean temperature inside that region, as a function of datetime.


src/plopp/plotting/inspector.py
```diff
@@ -22,46 +24,44 @@ def _to_bin_edges(da: sc.DataArray, dim: str) -> sc.DataArray:
...
 ) -> sc.DataArray:
-    from matplotlib.path import Path
-
-    xdim = poly['x']['dim']
-    ydim = poly['y']['dim']
-    x = _coord_to_centers(da, xdim)
-    y = _coord_to_centers(da, ydim)
-    vx = poly['x']['value'].to(unit=x.unit).values
-    vy = poly['y']['value'].to(unit=y.unit).values
+    vx = poly['x']['value'].values
```
nvaytet: I don't think we need to convert the unit here? By construction, the xy locations returned by the polygon info are in the correct canvas units.

nvaytet:
**Update:**

Instead of removing the `operation` or only allowing it for `point` mode, we apply the same operation along the x/y dims inside a selected region.
I think this makes sense: in the 1d plot, we would want to see what is the mean temperature inside that region, as a function of datetime.

src/plopp/plotting/inspector.py
```diff
@@ -32,36 +43,51 @@ def _apply_op(da: sc.DataArray, op: str, dim: str) -> sc.DataArray:
...
+        # If there is a 2D coordinate in the data, we need to slice the other dimension
+        # first. The assumption here is that there would only be one multi-dimensional
+        # coordinate in a given DataArray (which is very likely the case).
+        if da.coords[y['dim']].ndim > 1:
+            return da[x['dim'], x['value']][y['dim'], y['value']]
+        else:
+            return da[y['dim'], y['value']][x['dim'], x['value']]
+    except IndexError:
+        # If the index is out of bounds, return an empty DataArray
+        return sc.full_like(da[y['dim'], 0][x['dim'], 0], value=np.nan, dtype=float)
```
jokasimr: I didn't really understand this, what situation are we dealing with here?

src/plopp/graphics/bbox.py
```diff
@@ -107,6 +107,9 @@ def axis_bounds(
     pad:
         Whether to pad the limits.
     """
-    values = fix_empty_range(find_limits(x, scale=scale, pad=pad))
+    try:
+        values = fix_empty_range(find_limits(x, scale=scale, pad=pad))
+    except ValueError:
+        return dict.fromkeys(keys, None)
```
jokasimr: I guess this is what fixes the `numpy` issue, what was the problem?

src/plopp/plotting/inspector.py
jokasimr: I see now what you mean :+1:

PR #531, opened 2026-02-19
Include upper bound in 3d view V-cut

nvaytet:
The value-based (V) cut in 3D views was leaving out the upper bound.
Users would expect the upper bound given by the slider handle to be included in the selection (see also #526).

Example: data containing all zeros except one value = 100: the max value is left out of the selection
Screenshot 2026-02-19 at 12 39 03

We fix this by using `<=` instead of `<` in our selection.


celinedurniak (approved):
This change fixed the issue with the display as shown in the comments.

PR #550, opened 2026-04-07
docs: move to remote unzip on new deno deploy

jokasimr:


jokasimr:
Oh I thought Plopp was not using the copier template. But I see now that it is. Then we don't need to change this separately here.

PR #551, opened 2026-04-09
Speedup rectangle mode for inspector plot

nvaytet:
We use slicing for rectangle tool instead of masking like a polygon.
This yields a very large speedup.
Try out this on `main` and in this `speedup-rectangle-inspector` branch:

```Py
import plopp as pp
from plopp.data.examples import clusters3d

N = 512
da = clusters3d(nclusters=500, seed=12).hist(z=N, y=N, x=N)

pp.inspector(da, dim='z', mode='rectangle', logc=True)
```

### DIFF
--- a/src/plopp/plotting/inspector.py
+++ b/src/plopp/plotting/inspector.py
@@ -202,6 +202,13 @@ def inspector(
     Almost all the arguments for plot customization apply to the two-dimensional image
     (unless specified).
 
+    In rectangle mode, if one of the edges of the rectangle lies exactly on a bin edge
+    between two data points, the selected region does not include the data on the
+    outside of the rectangle, even though the edge is touching the rectangle.
+
+    In polygon mode, only data whose bin centers are inside the polygon are included in
+    the selected region.
+
     Parameters
     ----------
     obj:
--- a/src/plopp/plotting/inspector.py
+++ b/src/plopp/plotting/inspector.py
@@ -60,6 +60,42 @@ def _slice_xy(da: sc.DataArray, xy: dict[str, dict[str, int]]) -> sc.DataArray:
         return sc.full_like(da[y['dim'], 0][x['dim'], 0], value=np.nan, dtype=float)
 
 
+def _slice_rectangular_region(da: sc.DataArray, rect: dict, op: str) -> sc.DataArray:
+    x = rect['x']
+    y = rect['y']
+    xmin, xmax = x['value'].min(), x['value'].max()
+    ymin, ymax = y['value'].min(), y['value'].max()
+    try:
+        # If there is a 2D coordinate in the data, we need to slice the other dimension
+        # first, as trying to slice a 2D coordinate using label-based indexing raises an
+        # error in Scipp. After slicing the other dimension, the 2D coordinate will be
+        # 1D and can be sliced normally using label-based indexing.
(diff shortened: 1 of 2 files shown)


src/plopp/plotting/inspector.py
```diff
@@ -60,6 +60,42 @@ def _slice_xy(da: sc.DataArray, xy: dict[str, dict[str, int]]) -> sc.DataArray:
         return sc.full_like(da[y['dim'], 0][x['dim'], 0], value=np.nan, dtype=float)
 
 
+def _slice_rectangular_region(da: sc.DataArray, rect: dict, op: str) -> sc.DataArray:
+    x = rect['x']
+    y = rect['y']
+    xmin, xmax = x['value'].min(), x['value'].max()
+    ymin, ymax = y['value'].min(), y['value'].max()
```
SimonHeybrock, line 67: I presume the exclusive bounds are intentional?

nvaytet: Do you mean that if we have a `max` value that falls exactly on a bin edge, the bin to the right of that won't be included?
I think that's it what we want?
Screenshot_20260410_103524

If I draw a rectangle exactly from 15 to 20, then I want the cells from the left edge at 15 to the right edge at 20 to be included.
I don't want the cell with a right edge at 15 and the one with left edge at 20 to be included. Right?

SimonHeybrock: I don't know. We have made that decision for slicing, but I am not sure the same is valid for a UI and drawing a rectangle &mdash; a user might expect both bounds to be inclusive?

tests/plotting/inspector_test.py
```diff
@@ -420,8 +420,8 @@ def test_rectangle_mode():
 
     # This rectangle should select the bottom left corner of the data.
     # Closing the rectangle by repeating the first point.
-    x = [-1, 21]
-    y = [-1, 310]
+    x = [-1, 19]
```
nvaytet, line 423: This indicates a slight behaviour change:
before a cell was only included if its center was in the rectangle, now it's as soon as one edge is in the rectangle.
This makes little difference for data with a large resolution.

Which is the correct one is difficult to decide. When the operation we perform at the end is a sum, we would probably want to just rebin at the exact position of the rectangle edge, to include a fraction of an incomplete bin.
But if the operation is something else, it is not clear what should actually be done.

I think it's fine for now, it's probably what most users would expect.
If some fine tuning is needed for a specific application we can either revisit or make a custom selection function and inject that in the inspector (somehow).

src/plopp/plotting/inspector.py
nvaytet: I'm quite sure this is what the users will expect.

In any case, I decided to just document the behaviour in the docstring with
```
    In rectangle mode, if any part of a data pixel lies inside the rectangle, the whole
    pixel is included in the selected region (as opposed to computing the overlap area).
    This is because it is not always possible to know how the bin fraction should be
    included, depending on the reduction operation which is applied to the data (sum,
    mean, etc).
    If one of the edges of the rectangle lies exactly on a bin edge
    between two data points, the selected region does not include the data on the
    outside of the rectangle, even though the edge is touching the rectangle.

    In polygon mode, only data whose bin centers are inside the polygon are included in
    the selected region.
```

SimonHeybrock (approved):

PR #552, opened 2026-04-11
Slicer inspector

### LINKED ISSUE #15: Feature Request: Adding slicer to the inspector plot.

nvaytet:
Combining the range slicer with the inspector plot:
```Py
%matplotlib widget
import plopp as pp
from plopp.data.examples import clusters3d

N = 256
da = clusters3d(nclusters=500, seed=12).hist(z=N, y=N, x=N)
pp.inspector(da, dim='z', mode='rectangle', logc=True)
```
Screenshot_20260411_104410

Fixes https://github.com/scipp/ess/issues/15

### DIFF
--- a/docs/plotting/inspector-plot.ipynb
+++ b/docs/plotting/inspector-plot.ipynb
@@ -213,56 +213,12 @@
     "it is sometimes preferable to only update the target figure when we have stopped dragging or resizing one of the rectangles/polygons,\n",
     "instead of every step of the way.\n",
     "\n",
-    "This can be done using the `continuous_update` argument:"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "14",
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "p = pp.inspector(\n",
-    "    da,\n",
-    "    dim='z',\n",
-    "    orientation='vertical',\n",
-    "    logc=True,\n",
-    "    mode='rectangle',\n",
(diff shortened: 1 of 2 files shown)


docs/plotting/inspector-plot.ipynb
```diff
@@ -213,56 +213,12 @@
...
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "16",
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "p"
+    "This can be done using `continuous_update=False` in the argument list."
```
nvaytet, line 216: We don't need to render yet another figure just to illustrate the `continuous_update` arg. We reduce a little docs size bloat.

jl-wynen:
> `pp.inspector(da, dim='z', mode='rectangle', logc=True)`

You set `logc=True` in your example, should this also set `norm=log` for the 1D plot? Those to axes show the same quantity, right?

jl-wynen (approved):

nvaytet:
See discussion in https://github.com/scipp/plopp/pull/513#issuecomment-3784583977

It's always up for debate, but the current behaviour is intentional :wink:

PR #553, opened 2026-04-14
A more reusable Slicer

nvaytet:
We split the Slicer class into two parts:
- one that creates the slider widgets and slices the data (can access the result in the nodes stored in the `output` attribute
- one that creates the above + a figure 

The first one can then be used with other visualizations (e.g. the instrument view), while the second can be used by similar slicer views like the Superplot.

A quick example of something we can now do with this: update a html display from the slider
```Py
import plopp as pp
from plopp.plotting.slicer import Slicer
import ipywidgets as ipw

da = pp.data.data2d()

s = Slicer(da)

# Make a html widget
html = ipw.HTML()

# Function to display data array as html
def f(da):
    html.value = da._repr_html_()

pp.Node(f, da=s.output[0])

ipw.VBox([s.slider, html])
```
Screenshot_20260414_194358

This should hopefully be quite useful for integrating as part of custom GUI interfaces.

### DIFF
--- a/src/plopp/plotting/slicer.py
+++ b/src/plopp/plotting/slicer.py
@@ -119,8 +119,6 @@ class Slicer:
                 f'following dimensions were found: {[da.dims for da in dg.values()]}'
             )
 
-        # data_prototype = nodes[0]()
-        # dims = data_prototype.dims
         self.keep = _guess_keep_if_none(keep, dg.dims)
 
         if len(self.keep) == 0:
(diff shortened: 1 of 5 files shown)


src/plopp/plotting/slicer.py
```diff
@@ -83,33 +91,24 @@ class Slicer:
...
                 'mode to "single" to use the play button.'
             )
-        nodes = input_to_nodes(
-            obj,
-            processor=partial(preprocess, ignore_size=True, coords=coords),
-        )
 
-        dims = nodes[0]().dims
-        if keep is None:
-            keep = dims[-(2 if len(dims) > 2 else 1) :]
```
MridulS, line 109: Shouldn't we still keep these checks? Slicer can still accept `obj: PlottableMulti` as input?

MridulS (approved):

nvaytet:
I ended up using the new functionality in one of the gallery notebooks.
It shows well how it improves usability:

Code before:
```Py
slider = pw.SliceWidget(da, dims=['time'])
slider_node = pp.widget_node(slider)

slice_node = pw.slice_dims(data_array=da, slices=slider_node)

fig = pp.scatter3dfigure(slice_node, size=0.3, cbar=True)

pp.widgets.Box([fig, slider])
```

After:
```Py
from plopp.plotting.slicer import Slicer

sl = Slicer(da, keep='pixel')

fig = pp.scatter3dfigure(sl.output, size=0.3, cbar=True)

fig.bottom_bar.add(sl.slider)

fig
```

PR #559, opened 2026-04-27
Support updates to positions of scatter3d points

### LINKED ISSUE #495: Scatter3d plot: positions do not update
Only colors update.
This is good for performance, but counter-intuitive.

We need to find a way that does not impact performance as we are already at the performance limit for some of the instrument views.

nvaytet:
Alternative to #484 

Much of the changes are similar, but things are written in a simpler manner here, and we do not check for equality of coordinates (as this took time).

Updating the positions is between 3 and 10 times faster than updating the colors.
The gain of removing the unexpected behaviour that positions don't update (only colors used to update) are worth the cost, I think.

I tried looking at the Dream instrument view again and the loss of performance was not noticeable, contrary to what I wrote in https://github.com/scipp/plopp/pull/484#issuecomment-3274422255
Possibly because of performance updates done on the instrument view side itself? (https://github.com/scipp/scippneutron/pull/691)

Fixes #495

### DIFF
--- a/src/plopp/backends/pythreejs/scatter3d.py
+++ b/src/plopp/backends/pythreejs/scatter3d.py
@@ -132,7 +132,7 @@ class Scatter3d:
                         if self._color is None
                         else self._color
                     ),
-                    dtype='float32'
+                    dtype='float32',
                 ),
                 (self._data.coords[self._x].shape[0], 3),
             )
--- a/src/plopp/backends/pythreejs/scatter3d.py
+++ b/src/plopp/backends/pythreejs/scatter3d.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
+import time
 import uuid
 from typing import Literal
 
@@ -219,6 +220,7 @@ class Scatter3d:
         new_values:
             New data to update the point cloud values from.
         """
+        start = time.time()
         check_ndim(new_values, ndim=1, origin='Scatter3d')
         needs_redraw = new_values.shape != self._data.shape
         self._data = new_values
@@ -245,11 +247,16 @@ class Scatter3d:
             self._canvas.add(self.points)
 
         else:
+            here = time.time()
             if not self._static_positions:
                 self.position = self._make_positions()
+            print(f'Updating positions took {(time.time() - here) * 1000:.3f} ms')
+            here = time.time()
             if not self._static_colors and self._colormapper is not None:
                 self.color = self._make_colors()
+            print(f'Updating colors took {(time.time() - here) * 1000:.3f} ms')
 
+        print(f'Updating the scatter artist took {(time.time() - start) * 1000:.3f} ms')
         # if not self._static_positions:
         #     self.p
 
--- a/src/plopp/backends/pythreejs/scatter3d.py
+++ b/src/plopp/backends/pythreejs/scatter3d.py
@@ -65,6 +65,8 @@ class Scatter3d:
         artist_number: int = 0,
         opacity: float = 1.0,
         pixel_size: sc.Variable | float | None = None,
+        static_positions: bool = False,
+        static_colors: bool = False,
         mask_color: str | None = None,
     ):
         check_ndim(data, ndim=1, origin='Scatter3d')
@@ -75,6 +77,11 @@ class Scatter3d:
         self._x = x
         self._y = y
         self._z = z
+        self._static_positions = static_positions
+        self._static_colors = static_colors
+        self._color = color
+        self._artist_number = artist_number
+        # self._opacity = opacity
 
         # TODO: remove pixel_size in the next release
         self._size = size if pixel_size is None else pixel_size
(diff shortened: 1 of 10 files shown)


src/plopp/widgets/clip3d.py
```diff
@@ -479,6 +486,23 @@ def update_controls(self):
         opacity = self.opacity.value if at_least_one_cut else 1.0
         self._set_opacity({'new': opacity})
 
+        if (not self._nodes) and at_least_one_cut:
```
nvaytet, line 496: We change the mechanism here: instead of grabbing the data from the nodes in the view, we add children nodes to the data which is the input to the view.

This means that automatically, if the input data changes, the data in the cuts will also update.
We ca now do this because updating a scatter point with different number of points is now allowed (when we update the cut range, the number of selected points changes).

nvaytet:
Assigning you @jl-wynen as you already reviewed the older PR, and changes here are pretty similar (though some things were fixed with the clip3d).

src/plopp/backends/pythreejs/scatter3d.py
```diff
@@ -157,9 +168,62 @@ def update(self, new_values):
...
+                geometry=self._make_geometry(
+                    positions=self._make_positions(), colors=self._make_colors()
+                ),
+                material=self.material,
+            )
+            # Delay the removal of the old points until after the new points have been
+            # created, to minimize a flickering effect in the plot.
+            self._canvas.remove(self.points)
+            self.points = new_points
+            self._canvas.add(self.points)
```
jl-wynen, line 186: Could you further change this?

```suggestion
            self._canvas.add(self.points)
            self.points = new_points
            self._canvas.remove(self.points)
```

src/plopp/backends/pythreejs/scatter3d.py
```diff
@@ -137,17 +156,9 @@ def notify_artist(self, message: str) -> None:
...
-    def _update_colors(self):
-        """
-        Set the point cloud's rgba colors:
-        """
-        self.geometry.attributes["color"].array = self._colormapper.rgba(self.data)[
-            ..., :3
-        ].astype('float32')
-
-    def update(self, new_values):
+    def update(self, new_values: sc.DataArray) -> None:
```
jl-wynen, line 162: Is it feasible to have a simpler `update_values` method that only updates the colours and that gets called by nodes that do not change positions?

src/plopp/backends/pythreejs/scatter3d.py
```diff
@@ -92,41 +94,58 @@ def __init__(
...
+            return np.broadcast_to(
+                np.array(
+                    to_rgb(
+                        f'C{self._artist_number}'
+                        if self._color is None
+                        else self._color
+                    )
+                ),
+                (self._data.coords[self._x].shape[0], 3),
+            ).astype('float32')
```
jl-wynen:
```suggestion
            return np.broadcast_to(
                np.array(
                    to_rgb(
                        f'C{self._artist_number}'
                        if self._color is None
                        else self._color
                    ),
                    dtype='float32'
                ),
                (self._data.coords[self._x].shape[0], 3),
            )
```
To avoid copying the larger array after `broadcast`. If it's ok to have an array with 0-strides.

src/plopp/widgets/clip3d.py
```diff
@@ -327,7 +327,7 @@ def __init__(self, fig: BaseFig):
 
         self.tabs = ipw.Tab(layout={'width': '41.6em'})
         self._original_nodes = list(self._view.graph_nodes.values())
-        self._nodes = {}
+        # self._nodes = {}
```
jl-wynen:
```suggestion
```

src/plopp/widgets/clip3d.py
```diff
@@ -479,6 +480,23 @@ def update_controls(self):
...
         self._set_opacity({'new': opacity})
 
+        if (not self._nodes) and at_least_one_cut:
+            for n in self._original_nodes:
+                self._nodes[n.id] = Node(
+                    self._select_subset, da=n, cuts=self._cut_info_node
+                )
+                self._nodes[n.id].add_view(self._view)
+
+        if self._nodes and (not at_least_one_cut):
```
jl-wynen, line 503: Why `not at_least_one_cut` here? Don't you want to remove old nodes regardless of whether there is a cut now?

src/plopp/backends/pythreejs/scatter3d.py
nvaytet: Not sure I get it, why would I add the current points again instead of the new points?

src/plopp/backends/pythreejs/scatter3d.py
nvaytet: Not really, the values you get from an update are just a full data array with new data, where both coordinates or data could change.
If we want to change this, then it would be a much more fundamental change in how the whole Plopp works...

src/plopp/widgets/clip3d.py
nvaytet: So the logic is the following:

To start with, we have just the original data nodes (just say one of them for simplicity, which is the most common case).
When a button is clicked to create a new cut, we add a node which selects a subset of data points from the original node.

After that, no matter how many new cuts we add on top, we are still using that same node, which performs a selection from the originaly node, combining all selection criteria from the different cuts. There are at most only 2 nodes in play at any time: the original node and the node that does the selection.

Once the last cut gets deleted, that is when we need to remove the additional ('selection') node.
So this needs to happen when the final cut has been deleted.

Does that make sense?

src/plopp/backends/pythreejs/scatter3d.py
jl-wynen: Ah, I missed that `self.points` changes in the middle.
My idea was to add and then remove to not have a blank canvas in case that causes problems. But does 3js rerender when removing points? Or only when you explicitly tell it to?

src/plopp/widgets/clip3d.py
jl-wynen: Yes, I think I get it now.

jl-wynen (approved):

src/plopp/widgets/clip3d.py
nvaytet: I added the explanation as a comment in the code

PR #561, opened 2026-04-30
Make Scipp a hard dependency of Plopp

nvaytet:
See #560 for context and discussion.

Plopp uses Scipp everywhere internally. It makes sense that it should be a hard requirement.
Until now, we had it as a soft dependency because of fears of ending up in a circular dependency issue where plopp needs scipp but scipp needs plopp (because it imports it and adds the `.plot()` methods to DataArray).

But as long as Plopp just remains a soft dependency of Scipp, solving environments should be fine.

- if you `pip install plopp`, you would get plopp and scipp (and skip scipp if it is already installed)
- if you `pip install scipp`, you get only scipp
- if you `pip install scipp plopp` you get both at once, as expected.

This will also make it easier to keep our [conda recipe](https://github.com/conda-forge/plopp-feedstock/blob/main/recipe/recipe.yaml) healthy and in-sync with the pyproject.toml here.


SimonHeybrock (approved):

MridulS (approved):
I'm okay with this too, we just need to be consistent :)

PR #562, opened 2026-05-08
fix: replace set_over with_extremes

### LINKED ISSUE #564: set_over in matplotlib is being deprecated
See, e.g., the nightly job here: https://github.com/scipp/plopp/actions/runs/26077705769/job/76672058232

```
PendingDeprecationWarning: The set_over function will be deprecated in a future version. Use cmap.with_extremes(over=...) or Colormap(over=...) instead.
```

jokasimr:
Fixes the nightly test failures here: https://github.com/scipp/plopp/actions/runs/25536713387/job/74953922444

I'm not sure this is exactly equivalent to the old implementation. 
For example, in some cases we create new objects instead of mutating, I assumed that doesn't make a difference here, but maybe it does?


src/plopp/graphics/colormapper.py
nvaytet: I think that because we now return new objects, we no longer need this copy here?

nvaytet (approved):

PR #568, opened 2026-05-27
Fix normalization in mean operations in inspector rectangle mode

nvaytet:
The normalization factor when computing mean was including too many dimensions, resulting in too small values in the output.
Because of the dimensions of the test data (100 in size), the values were a factor of 100 too small, and we did not notice because the numbers looked the same, just the number of decimals differed...


nvaytet:
Approved by @damskii9992
Merging.

PR #514, opened 2026-01-22
Add grid toggle button to toolbar

nvaytet:
Screenshot_20260122_141053

Fixes #512

Issue #203, opened 2023-04-24
Add a method to convert plots to dict?


nvaytet:
It may be useful to have the ability to convert plots to a dict that would fully specify the figure.
Just like Scipp has `.to_dict()`, we could have something that would convert the plot to a `dict` which would be interpretable and plotted using any plotting library, not just Matplotlib.

The dict would contain all the info needed to build the figure (non-exhaustive list):
- the number of axes
- the limits of the axes
- the axes labels
- the type of plot (lines or image or contours...)
- the data to plot
- the font sizes
- the figure title

A python dict is easily convertible to JSON, making sending the data somewhere else more easy.

nvaytet:
Sunyoung and I discussed the possibility of using something like this to send the data to the dashboard for the live view. Let plopp interpet the data array and make a figure, and then send it as a dict or json to the dashboard, which can then use matplotlib or plotly or something else to re-create the figure.

The idea was to not have to re-create the logic we already have in plopp (what to do if data you receive is 1d or 2d etc..) in the dashboard if we use a library that is not Matplotlib (e.g. plotly).

nvaytet:
See https://discuss.scientific-python.org/t/is-there-a-plotting-standard-api-in-python-like-the-array-and-dataframe-standard-apis/715

nvaytet:
I see two things we could implement:

1. Add a `as_json()` method to an existing figure
2. Add a new backend which would not create any graphics but just have json blobs. It would still implement things like a canvas on which you can set ranges and scales

Both would be useful.

PR #571, opened 2026-06-10
Remove plotly backend

nvaytet:
The `plotly` backend, activated via `pp.backends['2d'] = 'plotly'` is basically unused.
It only has limited support for 1d line plots.

In addition, it is slowing down addition of new features (e.g. #514 ) and adding to maintenance burden.
We remove it for these reasons.

Moving forwards, we should be careful not to have matplotlib specifics creep in to part of the codebase which should be plotting-library-agnostic.


jl-wynen:
> Moving forwards, we should be careful not to have matplotlib specifics creep in to part of the codebase which should be plotting-library-agnostic.

How relevant is this really? Can we commit to matplotlib fully and simplify the code?

nvaytet:
> How relevant is this really? Can we commit to matplotlib fully and simplify the code?

That is something we should maybe think about. I guess it could simplify quite a bunch? We should still keep some clean separation between 2d and 3d plots (where we still want to use pythreejs), but otherwise I'm guessing matplotlib will live as long as python lives?

nvaytet:
Conclusion from in-person discussion with the DRAM analysis team:

The plotly backend in plopp is not needed by any current package. The apps that use plotly do their plotting manually with the library directly, and the ones that use plopp make use of matplotlib.

There are some useful parts to Plopp other than just the displaying of figure (data preprocessing, slicing, and a graph of connected nodes to interact with widgets). This part may still be useful even if you are not interested in matplotlib.

We decided that we could create a mode where instead of creating a figure, we output a json blob which contains all the information required to make a plot. Then, the groups are free to use whatever library they want to plot from the info in the json. It would be like serializing the plots.

For the record, this idea was inspired by #203 .

MridulS (approved):

PR #576, opened 2026-06-22
ci: use same setup as in other lower bound tests

jokasimr:


.github/workflows/nightly_at_main_lower_bound.yml
```diff
@@ -31,7 +31,8 @@ jobs:
         ESS_PROTECTED_FILESTORE_PASSWORD: ${{ secrets.ESS_PROTECTED_FILESTORE_PASSWORD }}
     steps:
       - uses: actions/checkout@v6
-      - uses: astral-sh/setup-uv@v8
+      - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
```
nvaytet, line 34: Should this be a copier template thing instead of changing all repos one by one??

jokasimr: Yes you are right, that's better

PR #581, opened 2026-06-24
Add 'band' mode for errorbars

### LINKED ISSUE #580: Add a 'band' mode for displaying errorbars?
It is sometimes more useful to display errorbars as a band around the data line (usually with lower opacity), especially when there are many points displayed.

bands

We could add such a mode via a kwarg. What would the best syntax be? Re-purpose the existing `errorbars` arg to accept a string in addition to `True/False`?
```Py
da.plot(errorbars='band')
da.plot(errorbars='bars')  # <-- current behaviour
da.plot(errorbars=True)  # Makes the same as 'bars'
```

nvaytet:
We expand the scope of the `errorbars` argument from being just `True/False` to also accepting a string to control the mode for representing errorbars on 1d plots. This can now be a `'band'` which plots as:
Figure 1 (80)

Example:
```Py
import plopp as pp

da = pp.data.data1d(variances=True)
da.plot(errorbars='band')
```

Fixes #580

### DIFF
--- a/src/plopp/backends/common.py
+++ b/src/plopp/backends/common.py
@@ -103,7 +103,7 @@ def make_line_bbox(
         # Note: [str(data.dims)] is used to make a unique dim name.
         line_y = sc.DataArray(
             data=sc.concat(
-                [data.data - stddevs, data.data + stddevs], dim=[str(data.dims)]
+                [data.data - stddevs, data.data + stddevs], dim=str(data.dims)
             ),
             masks=data.masks,
         )
--- a/src/plopp/backends/common.py
+++ b/src/plopp/backends/common.py
@@ -55,11 +55,7 @@ def make_line_data(data: sc.DataArray, dim: str) -> dict:
     values = {'x': xvalues, 'y': yvalues}
     mask = {'x': xvalues, 'y': np.full(y.shape, np.nan), 'visible': False}
     if data.variances is not None:
-        error = {
-            'x': xvalues,  # np.asarray(sc.midpoints(x).values) if hist else xvalues,
-            'y': yvalues,
-            'e': np.asarray(sc.stddevs(y).values),
-        }
+        error = {'x': xvalues, 'y': yvalues, 'e': np.asarray(sc.stddevs(y).values)}
     if len(data.masks):
         one_mask = np.asarray(merge_masks(data.masks).values)
         mask = {
--- a/src/plopp/backends/common.py
+++ b/src/plopp/backends/common.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 
-import uuid
 from typing import Literal
 
 import numpy as np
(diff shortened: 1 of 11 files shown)


src/plopp/backends/matplotlib/line.py
```diff
@@ -20,6 +21,154 @@ def _to_float(x):
...
+        elif self._mode == ErrorbarMode.bar:
+            if hist:
+                x = 0.5 * (x[1:] + x[:-1])  # Use bin centers for error bars
+            self._artist = ax.errorbar(
+                x, y, yerr=e, color=color, zorder=zorder, fmt="none"
+            )
+        else:
+            raise ValueError(f"Invalid errorbar mode: {mode}")
+
+    def update(self, x: np.ndarray, y: np.ndarray, e: np.ndarray, hist: bool) -> None:
```
jl-wynen, line 73: There is some repeated code in this function and `__init__`. Can you extract it?

src/plopp/backends/matplotlib/line.py
```diff
@@ -20,6 +21,154 @@ def _to_float(x):
...
+                yme = np.concatenate([yme[0:1], yme])
+                ype = np.concatenate([ype[0:1], ype])
+                extra_arg = {"step": "pre"}
+            color = self._artist.get_facecolor()[0]
+            alpha = self._artist.get_alpha()
+            zorder = self._artist.get_zorder()
+            self._artist.remove()
+            self._artist = self._ax.fill_between(
+                x, yme, ype, color=color, alpha=alpha, zorder=zorder, **extra_arg
+            )
```
jl-wynen: I guess you checked and this artist cannot be modified in place?

src/plopp/backends/common.py
```diff
@@ -57,7 +56,7 @@ def make_line_data(data: sc.DataArray, dim: str) -> dict:
     mask = {'x': xvalues, 'y': np.full(y.shape, np.nan), 'visible': False}
     if data.variances is not None:
         error = {
-            'x': np.asarray(sc.midpoints(x).values) if hist else xvalues,
+            'x': xvalues,  # np.asarray(sc.midpoints(x).values) if hist else xvalues,
```
jl-wynen: Should this commented code be removed?

src/plopp/backends/matplotlib/line.py
nvaytet: You can but it gets annoying when you have to handle both the line and step case. In the line case, the vertices are made of `len(x)*2 + 3` points (don't ask why +3) but in the step case it's `len(x)*4 + 3`.
To replace the vertices you have to use a mix of ordered points and points in reverse order, and with the step you have to insert some np.repeat in there as well.

I decided for the sake of simplicity to just remove and re-create. Performance seems fine on what I tested (although I could have tested on larger data; i will do that now just to make sure).

nvaytet: Turns out it is noticeably slower with larger data :disappointed:. I will try to update in-place.

src/plopp/backends/matplotlib/line.py
```diff
@@ -20,6 +21,171 @@ def _to_float(x):
...
+                    alpha=self._artist.get_alpha(),
+                    hist=hist,
+                )
+            else:
+                if hist:
+                    x2 = np.repeat(x, 2)
+                    xverts = np.concatenate([x[0:1], x2, x2[::-1][1:], x[0:1]])
+                    a = np.repeat(yme, 2)
+                    b = np.repeat(ype, 2)[::-1]
+                    yverts = np.concatenate([[b[-1]], a[0:1], a, b[0:1], b, b[-2:]])
```
jl-wynen, line 100: This does indeed look complicated. But it seems to make sense.

jl-wynen (approved):

src/plopp/backends/matplotlib/line.py
nvaytet: I hope the interface/number of vertices doesn't randomly change when they release a new matplotlib...

jl-wynen (approved):

PR #584, opened 2026-07-01
Make blank plot instead of raising on all nan

nvaytet:
Sometimes you have a slice of data that contains all NaNs, but you still want to inspect the rest of the data with e.g. the `slicer`.

Example:
```Py
import plopp as pp
%matplotlib widget

da = pp.data.data3d()
da['z', 0].values[...] = np.nan
pp.slicer(da)
```

This would raise with a message saying it cannot compute limits because data is all NaN.
Now we instead return a NaN range, which get converted to a null range [0, 0].
The plot is just blank instead. Moving the slider can still be used to view other healthy slices.

### DIFF
--- a/src/plopp/graphics/bbox.py
+++ b/src/plopp/graphics/bbox.py
@@ -7,6 +7,7 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Literal
 
+import numpy as np
 import scipp as sc
 
 from ..core.limits import find_limits, fix_empty_range
@@ -108,7 +109,11 @@ def axis_bounds(
         Whether to pad the limits.
     """
     try:
-        values = fix_empty_range(find_limits(x, scale=scale, pad=pad))
+        values = find_limits(x, scale=scale, pad=pad)
+        if np.isnan(values[0].value) or np.isnan(values[1].value):
+            return dict.fromkeys(keys, None)
+        else:
+            values = fix_empty_range(values)
     except ValueError:
         return dict.fromkeys(keys, None)
     bounds = dict(zip(keys, (val.value for val in values), strict=True))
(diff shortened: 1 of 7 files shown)


src/plopp/graphics/bbox.py
```diff
@@ -108,7 +109,11 @@ def axis_bounds(
         Whether to pad the limits.
     """
     try:
-        values = fix_empty_range(find_limits(x, scale=scale, pad=pad))
+        values = find_limits(x, scale=scale, pad=pad)
+        if np.isnan(values[0].value) or np.isnan(values[1].value):
+            return dict.fromkeys(keys, None)
```
jokasimr, line 114: If this is really meant to be treated equivalently to the case that raises ValueError then it could make sense to raise ValueError here too instead of returning the dict, so that we avoid duplicating that logic.

jokasimr (approved):

src/plopp/graphics/bbox.py
nvaytet: I think here we want to make a bbox with None, so that it gets ignored.

Say you plot 2 lines, and one of them has all NaN. I would expect to just see one line instead of having an error raised. You might then wonder why you have just one line, and in that case raising an error may be safer. But in the case of interactive plots, where one slice somewhere may have all nans and the others are fine, I think it makes sense to not show the data instead of raising.

jokasimr: I didn't mean to raise an error that is visible to the user. I meant that the `ValueError` will be catched by the surrounding try/catch, and then we return the same `dict.fromkeys` anyway but with less duplication.

nvaytet: So what you meant was:
```Py
        values = find_limits(x, scale=scale, pad=pad)
        if np.isnan(values[0].value) or np.isnan(values[1].value):
            raise ValueError("NaN limits")
```
?

jokasimr: Yes

Issue #596, opened 2026-07-31
Matplotlib math text parsing error with Python 3.13 when using plopp.slicer


jnianias:
When using plopp.slicer in a conda environment created from the official scipp.yml file and Python 3.13.5, Matplotlib fails to render scientific notation tick labels (e.g., 10^{0}) with a ParseException. I managed to resolve the issue by downgrading Matplotlib to version 3.9.4, originally was running 3.11.1.

Steps to Reproduce:

Using Python 3.13.5, install scipp via conda following the [official instructions](https://scipp.github.io/getting-started/installation.html).

Code:

```
import plopp as pp
import scipp as sc

%matplotlib widget

filename = 'https://public.esss.dk/groups/scipp/scipp/1/data_dream_diamond.zip'

ds = sc.io.load_csv(
    filename,
    sep='\t',
    header_parser='bracket',
)

table = sc.DataArray(
    sc.ones(sizes=ds.sizes, unit='counts'),
    coords={name:col.data for name, col in ds.items()},
)

binned_xyz = table.bin(z_pos=31, y_pos=31, x_pos=31)  # 31**3 bins
pp.slicer(
    binned_xyz['z_pos', 20:].hist(), keep=['x_pos', 'y_pos'], norm='log', aspect='equal'
)
```

Full traceback:

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[16], line 3
      1 # Multi-dimensional spatial binning
      2 binned_xyz = table.bin(z_pos=31, y_pos=31, x_pos=31)  # 31**3 bins
----> 3 pp.slicer(
      4     binned_xyz['z_pos', 20:].hist(), keep=['x_pos', 'y_pos'], norm='log', aspect='equal'
      5 )

File [~/miniconda3/envs/scipp/lib/python3.13/site-packages/plopp/plotting/slicer.py:162](http://localhost:8889/home/james/miniconda3/envs/scipp/lib/python3.13/site-packages/plopp/plotting/slicer.py#line=161), in slicer(obj, keep, autoscale, coords, vmin, vmax, cbar, **kwargs)
    124 def slicer(
    125     obj: PlottableMulti,
    126     *,
   (...)    133     **kwargs,
    134 ) -> FigureLike:
    135     """
    136     Plot a multi-dimensional object by slicing one or more of the dimensions.
    137     This will produce one slider per sliced dimension, below the figure.
[...]

Issue #499, opened 2025-11-10
`_ParseActionIndexError `  when displaying figure with `%matplotlib ipympl`


jokasimr:
Encountered

```python
File [~/.local/share/mamba/envs/essreduce/lib/python3.11/site-packages/pyparsing/core.py:307](http://localhost:8888/home/johannes/.local/share/mamba/envs/essreduce/lib/python3.11/site-packages/pyparsing/core.py#line=306), in _trim_arity.<locals>.wrapper(*args)
    305 except IndexError as ie:
    306     # wrap IndexErrors inside a _ParseActionIndexError
--> 307     raise _ParseActionIndexError(
    308         "IndexError raised in parse action", ie
    309     ).with_traceback(None)

_ParseActionIndexError: ('IndexError raised in parse action', IndexError('pop from empty list'))
```

when running this code

```python
sim = lut_wf.compute(SimulationResults)

def to_event_time_offset(sim):
    # Compute event_time_offset at the detector
    eto = (
        sim.time_of_arrival + ((Ltotal - sim.distance) / sim.speed).to(unit="us")
    ) % sc.scalar(1e6 / 14.0, unit="us")
    # Compute time-of-flight at the detector
    tof = (Ltotal / sim.speed).to(unit="us")
    return sc.DataArray(
        data=sim.weight,
        coords={"wavelength": sim.wavelength, "event_time_offset": eto, "tof": tof},
    )

events = to_event_time_offset(sim)
fig1 = events.hist(wavelength=300, event_time_offset=300).plot(norm="log")
fig2 = events.hist(tof=300, event_time_offset=300).plot(norm="log")
fig1 + fig2
```

in the [Dream tof notebook](https://scipp.github.io/essreduce/user-guide/tof/dream.html) in essreduce.

```
>>> matplotlib.__version__, ipympl.__version__, plopp.__version__
('3.9.0', '0.9.8', '25.10.0')
```

Running the same notebook without `ipympl` does not raise the error.

<details>
<summary>Full error message</summary>
<pre>
---------------------------------------------------------------------------
_ParseActionIndexError                    Traceback (most recent call last)
File [~/.local/share/mamba/envs/essreduce/lib/python3.11/site-packages/pyparsing/core.py:1306](http://localhost:8888/home/johannes/.local/share/mamba/envs/essreduce/lib/python3.11/site-packages/pyparsing/core.py#line=1305), in ParserElement.parse_string(self, instring, parse_all, parseAll)
   1305 try:
[...]

PR #597, opened 2026-07-31
Explain figure rendering failures caused by kernel subshells

SimonHeybrock:
Documents the JupyterLab subshell concurrency problem and its workarounds next to the installation instructions. Gives #596 the only answer plopp can give, and #499 a user-visible one.

JupyterLab >= 4.4 defaults to routing widget comm messages over kernel subshells, which ipykernel >= 7 services on their own threads. ipympl canvas draw/resize requests are then handled concurrently with cell execution, and Matplotlib is not thread-safe: a live canvas racing a figure build corrupts shared state. Symptoms are math text parse errors, blank figures, and kernel crashes, none of which point anywhere actionable. See [ipympl#610](https://github.com/matplotlib/ipympl/issues/610) for the analysis.

There is no fix on our side. [ipympl#621](https://github.com/matplotlib/ipympl/pull/621) locks Matplotlib's math text parser, which removes the parse errors, but the blank figures and crashes survive it: the frame path (`get_diff_image` reading `buffer_rgba()`) and `handle_resize` still run on the comm thread with no lockable API. Setting `commsOverSubshells` to `disabled` or pinning `ipykernel<7` are the only workarounds verified to be fully stable, so those are what we document.

An earlier version of this branch also chained the failure into an explanatory error at figure creation. That is dropped: it hooked the math text exception, which is exactly the symptom the ipympl lock removes, while what survives the lock raises nothing to hook.

## Test plan

- [ ] Docs build renders the new section.


nvaytet:
See also https://github.com/matplotlib/ipympl/pull/621

tests/backends/matplotlib/mpl_utils_test.py
nvaytet, line 1: I have my reservations as to how useful these tests are. We are trying to make a workaround for an issue that we hope will be fixed upstream.

Creating these fakes that mimic some private or undocumented properties of ipykernel feels like they could get out of date without us noticing.
We are also faking the error from the canvas, so the tests seem quite remote from the actual behaviour?

SimonHeybrock: Should we try monkey-patching as in the PR you linked instead?

nvaytet (approved):

PR #598, opened 2026-07-31
monkeypatch _mathtext.Parser.parse() with a threading.Lock()

nvaytet:
Copied from https://github.com/matplotlib/ipympl/pull/621

I don't know if this is the right place to put it, with the lazy_loader and everything. We may not want to do that work as soon as we import plopp?

I think we could potentially move it into the `backends/matplotlib/__init__.py` because the other places where matplotlib is used (e.g. making the colorbar for 3d scatter plots) don't make use of `ipympl`.

Either choice would lead to a somewhat possibly surprising behaviour for other plain MPL figures:

- if we keep it in top-level init, after importing plopp, it would fix other custom mpl figures
- in the backends/matplotlib/__init__, it would fix other figures after we have made a 1d or 2d figure with plopp

### DIFF
--- a/src/plopp/__init__.py
+++ b/src/plopp/__init__.py
@@ -44,4 +44,29 @@ from .backends import BackendManager
 
 backends = BackendManager()
 
+# Workaround for thread-safety issues in matplotlib's mathtext parser.
+# The _mathtext.Parser singleton has mutable state (_state_stack, etc.) that is
+# not protected against concurrent access from multiple threads. When ipympl is
+# used with ipykernel >= 7 (which introduces additional threads for message
+# routing), concurrent calls to the parser can corrupt this state, leading to
+# ParseException errors like 'Expected end of text, found "$"'.
+# See https://github.com/matplotlib/ipympl/issues/610
+try:
+    import functools
+    from threading import Lock
+
+    from matplotlib._mathtext import Parser as _MathTextInternalParser
+
+    _mathtext_parse_lock = Lock()
+    _original_mathtext_parse = _MathTextInternalParser.parse
+
+    @functools.wraps(_original_mathtext_parse)
+    def _thread_safe_mathtext_parse(self, *args, **kwargs):
+        with _mathtext_parse_lock:
+            return _original_mathtext_parse(self, *args, **kwargs)
+
+    _MathTextInternalParser.parse = _thread_safe_mathtext_parse
+except Exception:
+    pass
+
 del importlib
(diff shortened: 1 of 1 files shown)


SimonHeybrock:
Good find, I did not remember that PR.

I had Claude check that the lock genuinely does what it claims, and discussed whether we still need #597 (concluding to reduce it to docs-only)

---

(Claude's writing below): [...] four threads through `_mathtext.Parser.parse` with cache-defeating strings segfaults reliably here, and is clean with the lock in place. So this isn't a "does it work" objection.

Two things still make me want to skip it on our side.

It landed upstream: ipympl#621 was merged on 31 July, presumably after your ping. Latest release is still 0.10.0, so there is a real gap, but carrying a copy means double-locking once ipympl cuts a release, and remembering to remove it.

More importantly, the parser lock is only the first step of a chain. The comment on ipympl#610 serialized successive chokepoints with a shared RLock and watched the failure move each time:

- locking the parser removes the `ValueError`, and the race resurfaces as blank canvases and hard kernel crashes
- additionally locking `FigureCanvasAgg.draw` resurfaces `RuntimeError: ... did not call Figure.draw`
- additionally locking `print_figure` still leaves the frame path uncovered, since `get_diff_image()` reads `renderer.buffer_rgba()` and `handle_resize` mutates the figure on the comm thread, with no stable API to lock

On top of that, lock contention itself can starve the comm thread, so canvases stay blank until execution finishes anyway. I have not reproduced those downstream steps myself, only the parser race.

That is what tips it for me: we would trade a loud traceback for silent blank figures, which is harder to diagnose rather than easier.

On your placement question, the top-level `__init__` also costs us the lazy loading. `import plopp` is ~35 ms today and imports no matplotlib at all; pulling in `matplotlib._mathtext` drags `font_manager` along and takes it to ~200 ms.

I have reduced #597 to documentation only, pointing at `commsOverSubshells: disabled` and `ipykernel<7`, since those are the only workarounds reported as fully stable. Happy to be wrong here: if you can get a stable canvas with the lock under JupyterLab >= 4.4 and ipykernel >= 7, that would be worth reporting on ipympl#610.

nvaytet:
Maybe we should wait for ipympl to be released?

nvaytet:
> Maybe we should wait for ipympl to be released?

While the fix was merged into `ipympl`, I don't know when they plan to release. Their last one was 7 months ago.
I vote for getting this in now and removing it later when the fix is released upstream.

nvaytet:
> On your placement question, the top-level __init__ also costs us the lazy loading. import plopp is ~35 ms today and imports no matplotlib at all; pulling in matplotlib._mathtext drags font_manager along and takes it to ~200 ms.

I will leave it here for now and live with the slightly longer import time, and it seems to me the most sensible place to put it. I think giving users an unbroken experience is more important than a slightly slower import time.

PR #601, opened 2026-08-07
feat: add back and forward buttons

### LINKED ISSUE #587: Redo/undo buttons for widget
Especially for zooming in/out, it's often cumbersome to use home button and redo all the zooming again... cause you have to be very careful not to mess up again.

@bingli621 suggested

jokasimr:
Fixes #587

### DIFF
--- a/tests/backends/matplotlib/mpl_canvas_test.py
+++ b/tests/backends/matplotlib/mpl_canvas_test.py
@@ -173,13 +173,13 @@ def test_home_button_rescales_all_axes_sharing_a_figure():
 def test_back_and_forward_buttons_navigate_view_history(ndim):
     da = data_array(ndim=ndim)
     p = da.plot()
-    toolbar = p.canvas.fig.canvas.toolbar
+    mpl_toolbar = p.canvas.fig.canvas.toolbar
 
     original_range = p.canvas.xrange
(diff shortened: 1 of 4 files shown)


tests/backends/matplotlib/mpl_canvas_test.py
```diff
@@ -167,3 +167,22 @@ def test_home_button_rescales_all_axes_sharing_a_figure():
     p0.toolbar['home'].callback()
 
     assert p1.canvas.yrange == pytest.approx(expected)
+
+
+@pytest.mark.parametrize("ndim", [1, 2])
+def test_back_and_forward_buttons_navigate_view_history(ndim):
+    da = data_array(ndim=ndim)
+    p = da.plot()
+    toolbar = p.canvas.fig.canvas.toolbar
```
nvaytet: I got slightly confused here, because `toolbar` is the matplotlib toolbar. I guess you need that to be able to do `push_current()`. But I thought below we were then testing clicking buttons on the mpl toolbar. It's easy to miss that `toolbar` is the mpl one, and `p.toolbar` is our toolbar.

Can we either change the variable name or add a comment (or both)?

nvaytet (approved):

PR #604, opened 2026-08-10
feat: adds optional horizontal error bars

### LINKED ISSUE #295: Add error bars on coordinates
We currently only have vertical error bars.
We should add the possibility of plotting horizontal error bars also (e.g. variances on the coordinates)

jokasimr:
Fixes #295

### DIFF
--- a/src/plopp/backends/matplotlib/line.py
+++ b/src/plopp/backends/matplotlib/line.py
@@ -119,6 +119,8 @@ class Errorbars:
                 x = np.asarray(self._ax.convert_xunits(x))
                 y = np.asarray(self._ax.convert_yunits(y))
             else:
+                # For vertical errors, y is always numeric because Scipp only supports
+                # variances for numeric data. Only datetime x values need conversion.
                 x = _to_float(x)
             if hist and self._axis == 'y':
                 x = 0.5 * (x[1:] + x[:-1])  # Use bin centers for bars
--- a/src/plopp/backends/matplotlib/line.py
+++ b/src/plopp/backends/matplotlib/line.py
@@ -43,6 +43,7 @@ class Errorbars:
     def __init__(
         self,
         mode: Literal["band", "bar"],
+        axis: Literal['x', 'y'],
         ax: Axes,
         x: np.ndarray,
         y: np.ndarray,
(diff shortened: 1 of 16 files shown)


src/plopp/plotting/_plot.py
```diff
@@ -28,6 +28,7 @@ def plot(
     cmin: sc.Variable | float | None = None,
     coords: list[str] | None = None,
     errorbars: Literal['band', 'bar', True, False] = True,
+    errorbars_x: bool = False,
```
nvaytet: Not sure I like the interface here. It doesn't break existing code, but it feels weird to have a special `errorbars_x` next to `errorbars`.

Instead, I would suggest to add more possible values to `errorbar`: `xonly` and `yonly` (or something similar). The behaviour would then be the following:

- if there are both x and y errorbars, show them by default.
- show only y errorbars with `errorbars='yonly'`
- if `errorbars='bar'` is used and we have variances on the coord, just don't show the horizontal errorbars.

We can discuss about the last one, but I think it's probably the most predictable behaviour.

src/plopp/backends/matplotlib/line.py
```diff
@@ -43,6 +43,7 @@ class Errorbars:
     def __init__(
         self,
         mode: Literal["band", "bar"],
+        axis: Literal['x', 'y'],
```
nvaytet: I find the interface here strange, with both `mode` and `axis` args.
See other comment about having `xonly` and `ylonly` errorbar modes.
I would vote for always showing horizontal errorbars if present by default.

src/plopp/plotting/_plot.py
jokasimr: Hm, I don't think it makes sense to couple them so hard to  each other. To me the decision to have errorbars in x versus in y are separate decisions, and I'd prefer to not have one of those decisions coupled to the other.

But I'd be all for changing the name of the argument if you think it is unclear.

nvaytet: It's not that the name is unclear, it just feels strange to have `errorbars` and `errorbars_x`. It makes one think why not `errorbars_y`? Also, I would vote for showing them by default.

> To me the decision to have errorbars in x versus in y are separate decisions

You'd still have the option of turning one or the other off? I'm not sure I get the argument about coupling.

src/plopp/backends/matplotlib/line.py
```diff
@@ -53,26 +54,35 @@ def __init__(
...
-            if hist:
+            if hist and self._axis == 'y':
                 # Use bin centers for bars; We go via sc.midpoints as it handles
                 # datetime coordinates correctly.
                 x = np.asarray(sc.midpoints(sc.array(dims='x', values=x)).values)
             self._artist = ax.errorbar(
-                x, y, yerr=e, color=color, zorder=zorder, fmt="none"
+                x,
+                y,
+                xerr=e if self._axis == 'x' else None,
```
nvaytet, line 93: I don't think errorbars in x would work well here? We are using the center of the bins for display purposes, but it's a bit like cheating.
The variances are actually on the bin edges, meaning there is uncertainty about where the edges are located (which is conceptually weird).

I think I would just drop horizontal errorbars for plotting bin-edge data?

src/plopp/plotting/_plot.py
jokasimr: > I'm not sure I get the argument about coupling.

If I have an application with a button that says "turn on 'resolution' errorbars", then it is not enough to make that button alone control the `errorbar_x` argument. Instead the application has to check the state of the "turn on 'signal' errorbars" button and "turn on 'resolution' errorbars" button together, and based on them both decide what argument to pass when creating the figure.

If I want to change the errorbars in `y` or in `x` in isolation I can not just change the value of one flag, I have to first check what the current value is, and based on that determine how to change it.

That's the kind of stuff I mean.

jokasimr: > Also, I would vote for showing them by default.

That would be a quite significant behavioral change :/ and I don't think most people care much about the error bars in `x`. I think it would make many figures quite "noisy" and that's why I made it off by default.

But sure, we can turn them on by default.

src/plopp/backends/matplotlib/line.py
jokasimr: Agreed 👍 

> I think I would just drop horizontal errorbars for plotting bin-edge data?

Sounds good to me. We can always add support once we know how that case can be handled cleanly.

src/plopp/plotting/_plot.py
jokasimr: Changed it to match the interface you suggested.

PR #606, opened 2026-08-14
fix: inherit color from source artist

### LINKED ISSUE #384: Colors of clippings in 3d scatter plots do not match original colors
Example:
```Py
import plopp as pp
import scipp as sc

a = pp.data.scatter(1000)
b = a.copy()
b.coords['x'] += sc.scalar(60, unit='m')
pp.scatter3d({'a': a, 'b': b})
```
And then make a spatial clip in Y using the `clip3d` tool under the figure, the green and red opaque clipping should be blue and orange instead, respectively.
Screenshot at 2024-10-28 14-14-59

This is because the clippings are added as new data on the figure, and they automatically get given an incremented artist number, which cycles through the colors.

We don't see this when using `cbar=True` because all artists are using the same colormapper.

jokasimr:
Fixes #384

### DIFF
--- a/src/plopp/backends/pythreejs/scatter3d.py
+++ b/src/plopp/backends/pythreejs/scatter3d.py
@@ -203,12 +203,24 @@ class Scatter3d:
         return self.geometry.attributes['color'].array
 
     @color.setter
-    def color(self, val: np.ndarray):
+    def color(
+        self,
+        val: str
+        | tuple[float, float, float]
+        | tuple[float, float, float, float]
+        | np.ndarray,
+    ) -> None:
+        if isinstance(val, (str, tuple)) or val.ndim == 1:
+            color = val.tolist() if isinstance(val, np.ndarray) else val
+            if self._colormapper is None:
+                self._color = np.array(to_rgb(color), dtype='float32')
+            val = self._make_colors()
         self.geometry.attributes['color'].array = val
 
-    def _copy_color_from(self, other: 'Scatter3d') -> None:
-        self._color = other.color[0].copy()
-        self.color = self._make_colors()
+    @property
+    def single_color(self) -> np.ndarray | None:
+        """The single RGB color, or ``None`` if a colormapper is in use."""
+        return self._color
 
     @property
     def geometry(self) -> p3.BufferGeometry:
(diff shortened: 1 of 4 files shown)


src/plopp/backends/pythreejs/scatter3d.py
```diff
@@ -212,6 +206,10 @@ def color(self) -> np.ndarray:
     def color(self, val: np.ndarray):
         self.geometry.attributes['color'].array = val
```
nvaytet, line 218: I think it would be useful if we can fix something else here also. Currently, if you make a scatter
```Py
f = pp.scatter3d(da, pos='position', color='black', size=2)
artist, = f.view.artists.values()
```
if you try to update the color of the artist with
```Py
artist.color = 'red'
```
it fails, because the setter expects a numpy array with the correct size.

Can we support the above by doing more work here depending on input type/shape?
If it's a string or a tuple or a 1d numpy array (of length 3 or 4), then broadcast to the full array using `self._make_colors`, and just set the array otherwise?

`self._color` should then probably be updated in the setter also.

We could then maybe add an accessor to the `self._color` like `single_color` or `html_color` which would be `None` if a colormapper is in use (`html_color` is probably not great because we store it as a rgb, and `rgb_color` could be confusing because the current `.color` also returns rgb colors, just a large array of them).

Below in `clip3d.py`, we could then just have something like
```Py
self._view.artists[node.id].color = self._view.artists[source_id].single_color
```

src/plopp/widgets/clip3d.py
```diff
@@ -503,6 +503,11 @@ def update_controls(self):
             self._nodes.clear()
 
         self.update_state()
+        if self._view.colormapper is None:
```
nvaytet, line 506: Does it have to be done every time we update the controls or only when a cut is added? Meaning: can we avoid doing this also when a cut is removed?

**Edit:** looking more at the logic, I think this has to be called here.
Say we had 2 cuts: one in X and another in Y.
When we remove the Y cut, we are basically removing the old additional scatter that contained both X and Y selections, to now only have the selection for X. We are adding this new scatter and the colors need to be synced also.

nvaytet (approved):

PR #608, opened 2026-09-03
Fix plt.show()

### LINKED ISSUE #607: Matplotlib's plt.show() does not work
See https://github.com/scipp/plopp/pull/595#issuecomment-5131673827

In a script or ipython console:
```Py
import scipp as sc
f = sc.arange("x", 5.0).plot()
plt.show()
```
shows nothing, while using `f.show()` will show the figure (note that the figure may dissapear instantly again inside a script).

nvaytet:
Fixes #607

### DIFF
--- a/src/plopp/backends/matplotlib/utils.py
+++ b/src/plopp/backends/matplotlib/utils.py
@@ -90,7 +90,8 @@ def make_figure(*args, **kwargs) -> plt.Figure:
             _next_free_figure_num(), fig
         )
         # Register with pyplot global figure manager registry so plt.show() sees it.
-        Gcf.set_active(manager)
+        if not is_widget_backend():
+            Gcf.set_active(manager)
     return fig
 
 
--- a/src/plopp/backends/matplotlib/utils.py
+++ b/src/plopp/backends/matplotlib/utils.py
@@ -6,6 +6,7 @@ from typing import Literal
 
 import matplotlib as mpl
 import matplotlib.pyplot as plt
+from matplotlib._pylab_helpers import Gcf
 from matplotlib.lines import Line2D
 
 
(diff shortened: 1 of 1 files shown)


jokasimr:
Seems reasonable, but this is a test that fails on this branch but not on `main`:

```python
  import scipp as sc

  sc.arange("x", 5.0).plot()
  plt.close("all")  # Should succeed; PR 608 raises AttributeError.
```

nvaytet:
@jokasimr I decided to just use `plt.figure()` in cases where we are not in the notebook.
Hope this fixes everything

jokasimr:
Here's a test that AI found that passes on `main` but does not pass on this branch:

```python
  import plopp as pp
  import scipp as sc


  def test_tiling_leaves_no_temporary_figures():
      plt.switch_backend("WebAgg")
      left = pp.tiled(1, 1)
      right = sc.arange("x", 5.0).plot()
      plt.close("all")              # Close the inputs.
      plt.close((left + right).fig) # Create and close the result.
      assert plt.get_fignums() == []

```

I've verified manually that this actually leaves a non-closed figure that is displayed the next time the user calls `plt.show()`.

### The issue

  The issue: `left + right` creates an internal temporary figure. The PR registers that
  figure with pyplot without closing it afterward. Even after closing both inputs and the
  result, the temporary figure remains: the assertion fails with [1] == [], and plt.show()
  would display it.
  
  Difference between `main` and `plt-show-fix`:
  
 - `main`: Uses plt.Figure() and attaches a manager without registering it with pyplot.
    plt.show() ignores it—and also ignores the intended figures, which is the original
    bug.
- `plt-show-fix`: Uses plt.figure(), which registers every figure with pyplot. That
    includes the internal temporary figure. It stays registered after the result is
    closed, so plt.show() displays an unwanted extra plot.

### Possible fix

In `Tiled.__add__ (src/plopp/backends/matplotlib/tiled.py:139)`, close the temporary figure:

```python
  t = Tiled(1, 1)
  t[0, 0] = other
  plt.close(t.fig)  # Added
  other = t
```
  Closing removes the figure from pyplot and closes its window. Its contents remain
  available for copying into the final result.

  Apply the same line to the temporary wrappers in `__truediv__`, `hstack`, and `vstack`—six
  locations total.

PR #609, opened 2026-09-14
Option to disable log buttons

nvaytet:
Sometimes, when we have many subplots, the log buttons on every axes get in the way (and they also slow things down when the mouse enters/leaves the axes).
We add an option to disable them.


src/plopp/backends/matplotlib/canvas.py
```diff
@@ -256,26 +259,33 @@ def __init__(
...
-                ax=self.ax, label="logY", position=(-0.015, 0.98), **args
-            )
-
-            if self.cax is not None:
-                args = {"transform": self.cax.transAxes, "ha": "center"}
-                self._logc_button = CanvasToggleButton(
-                    ax=self.cax, label="log", position=(0.5, 0.98), va="top", **args
+            if not hide_log_buttons:
+                args = {"transform": self.ax.transAxes, "ha": "right", "va": "top"}
+                self._logx_button = CanvasToggleButton(
```
jl-wynen, line 264: Are the attributes on `self` optional now? How does the rest of the class cope with this?

src/plopp/widgets/hoverbutton.py
```diff
@@ -148,3 +148,57 @@ def on_fit_button_click(self, handler):
...
+
+      // Create SVG container
+      const svgContainer = document.createElement('div');
+      svgContainer.style.width = '100%';
+      svgContainer.style.lineHeight = '0';
+
+      // Function to update SVG
+      function updateSVG() {
+        const svgData = new TextDecoder().decode(model.get('svg_data'));
+        svgContainer.innerHTML = svgData;
```
jl-wynen, line 174: You are sending unvalidated HTML to the DOM here. The same in `HoverButtonWidget`. But I guess this is well isolated enough that it doesn't receive arbitrary inputs?

nvaytet: I guess I am? Suggestions for change?

src/plopp/backends/matplotlib/canvas.py
nvaytet: The diff is misleading, the `self._logx_button` are set to None further up.

src/plopp/widgets/hoverbutton.py
jl-wynen: To be completely safe, I think you would have to do a bunch of validation manually or get a 3rd party library. But using an image element like this should help:
```html
 <img src="data:image/svg+xml,{image-data}">
```
where `{image-data}` is the SVG source as a base64 encoded string.

Generally, this widget only receives inputs from Python, i.e., a process with system access. Python can do much worse than inject some malicious svg. So this might not be a concern.

nvaytet: Not sure I understood the conclusion here. Is it just to use the suggested image tag and that's it?

nvaytet: See update

jl-wynen (approved):

